PR #2, opened 2019-09-30
removed non-scientific name from scientific name

TheCodeSharman:
A branch for discussing issues with the inky key files

### DIFF
--- a/walta-taxonomy/walta/taxa.ink
+++ b/walta-taxonomy/walta/taxa.ink
@@ -946,7 +946,7 @@
 === attack_caddis_lowlander ===
 # taxonId: "166"
 # name: "Some Oecetis sp. (Leptoceridae) and Odontoceridae"
-# scientificName: [{"taxonomicLevel":"phylum","name":"Arthropoda"},{"taxonomicLevel":"subphylum","name":"Hexapoda"},{"taxonomicLevel":"class","name":"Insecta"},{"taxonomicLevel":"order","name":"Trichoptera"},{"taxonomicLevel":"family","name":"Leptoceridae"},{"taxonomicLevel":"genus","name":"Oecetis"},{"name":"Some Oecetis sp. (Leptoceridae) and Odontoceridae"}]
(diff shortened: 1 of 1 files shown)


walta-taxonomy/walta/taxa.ink
```diff
@@ -946,7 +946,7 @@
 === attack_caddis_lowlander ===
 # taxonId: "166"
 # name: "Some Oecetis sp. (Leptoceridae) and Odontoceridae"
-# scientificName: [{"taxonomicLevel":"phylum","name":"Arthropoda"},{"taxonomicLevel":"subphylum","name":"Hexapoda"},{"taxonomicLevel":"class","name":"Insecta"},{"taxonomicLevel":"order","name":"Trichoptera"},{"taxonomicLevel":"family","name":"Leptoceridae"},{"taxonomicLevel":"genus","name":"Oecetis"},{"name":"Some Oecetis sp. (Leptoceridae) and Odontoceridae"}]
```
TheCodeSharman: I've removed the "name" key because the app currently doesn't understand "name" as a scientific level. This results in a funny looking taxon details that includes "undefined" in the scientific levels.
Screenshot (30 Sep  2019 8_26_55 pm)

TheCodeSharman: This raises an interest point about field names, why is there 3 name fields? I currently only display commonName but there is also a "name" - when/where shoulkd I display this? Should it be removed.

TheCodeSharman: One other point. This file is humongous. It would be an easy task to split this into several files. Is there any reason we couldn't have a file per "order"? That might be a good break down. It would make finding work with the text much easier.

TheCodeSharman: Another issue, the names of the species are some times all lower case but other times camel case. I guess I could program a function to force camel case - but maybe cleaning them up is a good idea.

TheCodeSharman: Also might be a good idea to consistently use punctuation. e.g. "'U' bent midges" vs "U sthenids" seems inconsistent to me

TheCodeSharman: "undefined" is actually quite appropriate because the this category doesn't map well to the formal scientific calssification.

PR #11, opened 2020-06-23
Vector emergence map

TheCodeSharman:
This pull request is a work in progress:
 - add a Leaflet layer that efficiently pulls WFS vector data for the mayfly emergence map
 - this enables dynamic creation of emergence dates for the current window


TheCodeSharman:
We're abandoning this patch until we actually have the data to make this function properly - for the moment it's easier to use WMS SLD syle to render mayfly observations.

PR #198, opened 2026-04-23
WB-34: Implement new synchronise user interface in SampleHistory

TheCodeSharman:
## Summary

Adds a **Sync Feedback popup** to SampleHistory that shows live progress, captures the full log history (including syncs already running when the popup is opened), and handles success / error / offline states per the Figma design.

Also introduces the project's first **MVVM** layer (pure-JS ViewModel in `lib/logic/viewmodels/`, pure-JS domain model in `lib/models/`) so view logic is testable at Node speed and decoupled from Alloy.

Two final commits land a substantial **acceptance-test infrastructure overhaul** that the sync scenario depends on — see "Test infrastructure" below.

Trello: [WB-34](https://trello.com/c/jd0H0WWE/34)

## What's in this branch (8 commits)

1. **MVVM scaffolding** — `lib/logic/viewmodels/README.md` design doc, `SyncFeedback` VM skeleton, Mocha glob widened to `test/**/*_spec.js`.
2. **Sync lifecycle events** — new `SYNC_STARTED` / `SYNC_FINISHED` topics published by `SampleSync`; each `UPLOAD_PROGRESS` now carries a human-readable `message`.
3. **ViewModel state machine** — 5 states (`idle / syncing / success / error / offline`), log visibility, close / diagnostics events, log-cap.
4. **Alloy view + controller** — modal overlay, progress bar (teal / red), log pane, button row, on-device smoke spec.
5. **Sync Now button in SampleHistory** — explicit user-triggered popup.
6. **SyncStore domain model** — `SampleSync` now owns a `SyncStore` and is the source of truth for sync state. The VM is a pure projection of it.
7. **`walta://` deeplink + adb-direct Android launch** — see Test infrastructure below.
8. **`AppiumLauncher.setLocation()`** — encapsulates the GPS-fix shell-out so step defs don't reach into adb/xcrun directly.

## Test infrastructure

The sync acceptance scenario surfaced several latent issues in how the test rig launched the app and authenticated. Fixed in the last two commits:

- **`walta://login` URL scheme** for acceptance login. Replaces the launch-arg auto-login with an explicit `Given I am logged in as "<email>"` step that fires the deeplink. The scheme is registered in `tiapp.xml.template` (iOS `CFBundleURLTypes`, Android intent-filter on `.WaterbugActivity` with `launchMode="singleTask"`) and dispatched via a small action registry in `app/lib/UrlActions.js` (open for future programmatic test actions). Production-safe — login still requires valid credentials.
- **Per-launch config via `adb shell am start` directly.** uiautomator2's `mobile: startActivity optionalIntentArguments` is silently ignored; per-scenario re-launches were arriving without the mock-server URL extra and sync was hitting the production sandbox. `AppiumLauncher.launch()` now spawns adb directly on Android.
- **Mock server always-on** for acceptance runs. `@mockserver` tag retired (was only ever filtering 1 scenario). Mock now logs every request to `/tmp/mock-cerdi.log` for diagnosis, and the token-create stubs are `.many()` so multiple logins across scenarios work.
- **Explicit `Given the GPS has a fix` step** (was an implicit per-Before nudge that raced with `Ti.Geolocation` listener attach).
- **`AppiumLauncher.setLocation(lat, lng)`** — keeps platform-specific shell-out (adb / xcrun) in one place, unit-tested through `_spawn` injection.

State-reset between scenarios is plain `terminate + relaunch` — the heavier `pm clear` (Android) / `removeApp + installApp` (iOS) approaches were tried and either over-isolated (iOS sim destabilised) or duplicated platform-specific complexity. A cross-platform `walta://reset` deeplink action would be the natural replacement if sample-DB leakage starts to bite — not needed today.

## Known follow-up

- **WB-36**: [Pin vite to 4.x to restore LiveView fast-iteration workflow](https://trello.com/c/CYAJlfya). Discovered during manual verification — root `vite@8` (ESM-only) breaks `require('vite')` in `walta-app/vite.config.js`. Not a regression.
- **WB-43**: [Compose AppiumLauncher with Android/IosLauncher to remove duplicated launch/terminate](https://trello.com/c/7dk1UnW3). The dedicated `AndroidLauncher` / `IosSimulatorLauncher` already own adb/simctl shell-out for the build pipeline; AppiumLauncher reimplements it. Worth consolidating in a focused refactor — explicitly out of scope here.

## Test plan

- [x] `npx grunt unit-test-node` — **86 passing** (SyncStore + ViewModel + UrlActions + existing).
- [x] `npx grunt --platform=android --simulator acceptance-test` — **4 scenarios (4 passed)** locally.
- [x] `npx grunt --platform=ios --simulator acceptance-test` — **4 scenarios (4 passed)** locally.
- [x] CI green on commit `81ab1c45` (Android + iOS unit + acceptance, Node unit). Awaiting CI on `94f40cd3` (setLocation refactor).
- [ ] Manual smoke (post-WB-36, with liveview): SampleHistory → Sync Now → verify syncing / success / error / offline states visually.


walta-app/app/controllers/SyncFeedback.js
```diff
@@ -0,0 +1,57 @@
...
+
+var FILL_TEAL = "#26849c";
+var FILL_RED = "#c0392b";
+
+var vm = new SyncFeedbackViewModel({ syncController: SampleSync });
+
+function render(state) {
+    $.offlineMessage.visible = state.status === "offline";
+
+    var isError = state.status === "error";
```
TheCodeSharman: Can we move this logic into the `SyncFeebackViewModel`? The intent is to keep the controllers as thin as possible: they are only concerned with the plumbing not the content. 

The colour of the progress bar should be calculated iniside `SyncFeebackViewModel`, along with formatting the percent and adding the text. The view model should calculate and return: 
 - the colour of the progress bar
 - and the text value. 
 
 This keeps view logic inside the view model where it can be reused by other frameworks instead of inside the Titanium controller which couples it to Titanium.

PR #204, opened 2026-04-24
SPIKE: SyncFeedback VM as Alloy Model + XML data-binding

TheCodeSharman:
Spike to evaluate the alternative VM shape: make the ViewModel an **Alloy Model** (i.e. a `Backbone.Model` subclass at `app/models/syncFeedback.js`) so the view XML can bind attributes declaratively (`{syncFeedback.progressColor}`) instead of the controller driving updates via an explicit `render()` function.

Base is `task/wb-34-sync-feedback-ui` so the diff is purely the VM shape change.

## What changed

- **`walta-app/app/models/syncFeedback.js`** — new Alloy Model, in-memory (`properties` adapter), no SQL. `initialize(attrs, options)` accepts `{ syncController }` so Node tests can inject a fake; defaults to the shipped `SampleSync` singleton.
- **`walta-app/app/views/SyncFeedback.xml`** — `<Model src=\"syncFeedback\"/>` declaration + `{syncFeedback.<attr>}` bindings on every dynamic attribute.
- **`walta-app/app/controllers/SyncFeedback.js`** — no `render()`. Just wires button clicks to VM actions and forwards discrete events.
- **`walta-app/app/lib/logic/viewmodels/SyncFeedback.js`** — deleted (superseded by the Alloy Model).
- **`test/helpers/alloyModel.js`** — new helper that lets Node tests instantiate an Alloy Model without the Alloy runtime — applies `extendModel` to a fresh `Backbone.Model.extend({})` subclass.
- **`test/viewmodels/SyncFeedback_spec.js`** — rewritten to use `.get(attr)` / `.on('change:attr')` (Backbone) instead of property getters / `addListener`.

## Wrinkles surfaced by the spike

1. **Singleton semantics.** `Alloy.Models.syncFeedback` is a shared instance — state (e.g. `logVisible`) persists across popup opens. Feature or bug depending on taste.
2. **Don't `vm.dispose()` in controller cleanUp.** Disposal calls `vm.off()` which removes Alloy's generated binding listener AND the model's own internal `change:*` wiring set in `initialize()`. Since `initialize` only runs once per singleton, subsequent popups break.
3. **Initial render.** Alloy's binding listens to `change`/`fetch`/`destroy`. Since we don't `fetch`, the initial render never fires unless the controller calls `vm.trigger(\"change\")` after setup.
4. **Computed attributes must be materialised.** `progressColor` etc. can't be lazy getters — they must live in the attributes bag via `.set()` calls from a `_recomputeDerived` function that runs on every source-attr change.
5. **Underscore isn't available as a Node-resolvable module under Titanium**, so the model uses `Object.assign` instead of `_.extend`.

## Test status

- Node (`npx grunt unit-test-node`): **52 passing** (SyncFeedback VM + SyncStore + ChangeNotifier).
- Device (`npx grunt --platform=ios --simulator --liveview --grep=SyncFeedback unit-test`): **2 passing**.

## What to look at

- Does the thinner controller + declarative XML bindings feel enough nicer to adopt this shape for future VMs?
- Are the singleton semantics a blocker for any VM that needs per-popup state?
- Is the `_recomputeDerived` materialisation pattern acceptable compared to lazy getters?

Do **not merge** — this is a spike for design evaluation only.


TheCodeSharman:
This is worse than plain JS view model so closing this.

PR #216, opened 2026-04-30
Acceptance retry gate: skip retry when cucumber reports test failures

TheCodeSharman:
## Why

Extends the [WB-48](https://trello.com/c/iKBTKnlE) unit-test gating pattern to acceptance tests. Today the Android and iOS acceptance jobs retry once on **any** non-zero exit ([WB-54](https://trello.com/c/PZ5tnhi5)), which means a deterministic test failure burns ~15 min on iOS / ~7 min on Android re-running a suite that's guaranteed to fail again.

Surfaced today on PR #214 — Android acceptance failed at the close-popup step and the retry took a full second pass before reporting red.

## What

Cucumber-js exit codes:
- \`0\` — all scenarios passed
- \`1\` — ran to completion, some scenarios failed (deterministic)
- \`*\` — pre-launch / framework / infrastructure failure

Changes:

- \`Gruntfile.js\` cucumber task branches on the launcher's exit code:
  - \`0\` → \`done()\`
  - \`1\` → \`process.exit(2)\` (deterministic, no retry)
  - other → \`done(false)\` (infra flake, retry)

- \`ci.yml\` Android acceptance: replaces the inline \`||\` retry block with \`bash build-utils/run-unit-tests-with-retry.sh sh -c "preflight && acceptance"\`. Wrapper already keys off exit 2 from WB-48.

- \`ci.yml\` iOS acceptance: replaces the \`for attempt in 1 2\` loop with the same wrapper. iOS preflight retry left untouched — it's not cucumber, all failures there are infra-ish.

- \`build-utils/run-unit-tests-with-retry.sh\` header comments updated to reflect dual use. Filename retained; cosmetic rename is a follow-up.

The 5 existing wrapper unit tests already cover the exit-code contract.

## Test plan

- [x] \`npx grunt build-test\` — 67 passing locally.
- [ ] **CI on this PR**: expect all jobs green (acceptance suites pass on main today, no scenarios failing).
- [ ] **Real verification** comes when a deterministic acceptance failure lands — the failing job should report red on the **first attempt**, with no retry-cycle wasted.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -847,14 +847,20 @@ jobs:
           avd-name: Medium_Phone_API_36.1
           force-avd-creation: false
           emulator-options: -no-window -gpu swiftshader_indirect -noaudio -no-boot-anim -camera-back none
-          # Retry once on infrastructure flake (WB-54: AAPT
-          # `Theme.WaterbugApp not found` resource-linker glitch,
-          # Appium hiccups, etc.). Skip the retry if cucumber-js
-          # exited 1 → grunt exit 2 → wrapper exit 1: the failure is a
-          # deterministic test fail and a retry is just burning cycles.
-          # Same exit-code contract as the WB-48 unit-test gating.
-          # Wraps both preflight and acceptance so either phase being
-          # hit by an infra flake gets the second shot.
+          # Two phases — preflight then cucumber — each independently
+          # retry-wrapped, mirroring the iOS structure (preflight + Run
+          # iOS acceptance tests are separate steps there). The
+          # emulator-runner action runs each `script:` line via its own
+          # `sh -c`, so a failed preflight after retry doesn't
+          # short-circuit the cucumber phase — the cucumber phase will
+          # also fail and the overall step goes red. The wasted ~5 min
+          # is acceptable for the readability win in CI logs (each
+          # phase reports separately).
+          #
+          # Retry semantics on each phase: skip retry if grunt exits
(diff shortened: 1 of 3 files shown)


TheCodeSharman:
Pausing this PR — investigating the iOS acceptance failure showed that cucumber-js exit codes can't distinguish hook timeouts from scenario assertion failures, so the WB-54 fast-fail gate was misclassifying an XCTDaemon launch timeout as deterministic and not retrying.

Eliminating the hook flake at the source ([WB-67](https://trello.com/c/8v87ndxJ/67-replace-per-scenario-terminatelaunch-with-walta-reset-deeplink) — replace per-scenario terminate+launch with a `walta://reset` deeplink) is cheaper and more correct than the exit-code parser route. Once that lands and acceptance is stable, we can revisit whether this gate is still needed at all.

TheCodeSharman:
Superseded by [WB-67](https://trello.com/c/8v87ndxJ) (eliminates the per-scenario hook flake via `walta://reset` deeplink) and [WB-69](https://trello.com/c/TBD/) (moves per-suite-startup setup out of cucumber-js so launch failures are distinguishable from test failures by exit code alone — no output parsing needed).

The exit-code-only retry gate this PR tried to introduce can come back as a small follow-up once WB-69 lands.

TheCodeSharman:
(Correcting the WB-69 link in the close comment — Trello shortLink: https://trello.com/c/IjKvPWqd/69-move-test-infrastructure-setup-into-cucumberlauncher-distinguish-launch-failures-from-test-failures)

PR #226, opened 2026-05-06
WB-67: walta://reset deeplink for in-app per-scenario reset

TheCodeSharman:
## Summary
- New `walta://reset` deeplink action that, when fired into the running app, performs an in-app reset: clears auth tokens, wipes the local sample/taxa tables, re-instantiates `Alloy.Models`, and fires `Topics.HOME` to navigate back to the menu.
- Cucumber's per-scenario `Before` hook now fires the deeplink (`mobile: deepLink`) instead of `terminateApp + launch + foreground-poll`. That sequence was hitting the 60s default Before-hook timeout on iOS sim flakes (XCTDaemon "Timed out attempting to launch app") — see PR #216 for the failure that motivated this work.
- **Production safety**: the `reset` action is only registered when `UrlActions.create()` receives an `appReset` callback, and `index-app.js` only passes one when `Ti.App.deployType !== 'production'`. Release builds never know the action exists.

[Trello WB-67](https://trello.com/c/8v87ndxJ/67-replace-per-scenario-terminatelaunch-with-walta-reset-deeplink). Follow-up: [WB-68](https://trello.com/c/hqTjWkKg/68-encapsulate-device-prep-into-androidlauncher-iossimulatorlauncher-drop-redundant-beforeall-commands) covers moving the BeforeAll `adb`/`simctl` device-prep into the launcher classes.

## Why this approach over PR #216's exit-code gate

cucumber-js exits 1 for *both* assertion failures *and* hook failures — they're indistinguishable from the exit code alone. PR #216's WB-54 retry-gate misclassified an XCTDaemon hook timeout as a deterministic test failure and skipped retry. Eliminating the hook flake at the source (no slow per-scenario relaunch) is cheaper than parsing cucumber output to tell flakes from real failures.

## Test plan
- [x] `npx grunt unit-test-node` — 119 specs pass (new `dispatch(reset)` cases in `UrlActions_spec`)
- [x] `npx grunt build-test` — 67 specs pass
- [ ] CI green (iOS + Android acceptance, end-to-end)

### DIFF
--- a/features/support/cucumber.js
+++ b/features/support/cucumber.js
@@ -131,30 +131,31 @@ Before(async function () {
     this.isSimulator = global.isSimulator;
     setUpWorld(this);
     if (!global.first) {
-        // driver.reset() was removed in webdriverio v9 — terminate +
-        // relaunch is enough to restart the JS context and let the
-        // deeplink login re-authenticate. We deliberately do NOT wipe
-        // app data here: pm clear (Android) is fast but the iOS
-        // equivalent (removeApp + installApp) destabilises the
-        // simulator. Sample-DB leakage between scenarios is tolerable
-        // today; if it bites we'll reintroduce a cross-platform reset
-        // via a `walta://reset` deeplink action rather than per-platform
-        // hacks.
+        // WB-67: in-app reset via the `walta://reset` deeplink replaces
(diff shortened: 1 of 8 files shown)


TheCodeSharman (commented):
Looks - very quick review

PR #255, opened 2026-05-19
WB-92: Replace looks-same with macOS-arm64-safe image diff

TheCodeSharman:
## Summary

- **Fix:** swap `looks-same@10.0.1` + `images` for `pixelmatch` + `pngjs`. `looks-same` hangs indefinitely on macOS arm64 — its callback is never invoked, verified with a direct `node -e` against a real baseline PNG. iOS acceptance runs on macOS, so any pixel-diff step would stall the suite on Apple Silicon dev machines and CI. `pixelmatch`+`pngjs` are pure JS, zero native deps.
- **Bonus bug fixed:** `image-test.js:14` checked `path.extname(filePath) !== "png"` — but `path.extname()` returns `.png` with the leading dot, so the branch was always taken, routing every input through the `images` native PNG re-encode.
- **New unit coverage:** `test/image-test_spec.js` exercises `assertLooksSame` against real baseline PNGs (identity comparison + distinct-image comparison). Catches the hang in <1s instead of stalling cucumber. First Node-level coverage for `features/support/image-test.js`.
- Public API (`assertLooksSame`, `diffImages`) signatures unchanged, so existing callers don't move.

First slice of [Trello WB-92](https://trello.com/c/BItoj1Eo/92-replace-looks-same-with-macos-arm64-safe-image-diff) — and the slice is the whole card; nothing deferred. Unblocks WB-89 ("Photos missing on iOS"), whose test branch rebases on top.

Latent-bug note: the only existing `assertLooksSame` caller lives in `features/edit_samples.feature`, which is `@skip` at feature-level, so the hang never triggered in CI. Surfaced while planning WB-89.

## Test plan

- [x] `npx grunt unit-test-node` — passes (169 specs, including the new image-test_spec.js)
- [x] No active `require('looks-same')` or `require('images')` remain (verified via grep — sole remaining reference is inside a `/* */` block in `contract-tests/CerdiApi_spec.js`)
- [ ] CI green on main-equivalent runners


TheCodeSharman (commented):
Looks good to me

PR #261, opened 2026-05-20
WB-88: Inconsistent Sync — heal stale-container photo paths

TheCodeSharman:
## Summary
- **Root cause:** creature/site photos render blank on iOS after an app update because the DB stored absolute paths embedding the iOS data-container UUID (`…/Application/<UUID>/Documents/…`). That UUID changes across updates/reinstalls, so the file survives under a new container but the stored path resolves to nothing. A mix of old/new UUIDs across rows is exactly why *some* photos show and some don't — the "Inconsistent Sync" symptom.
- **Heal on read:** `PhotoUtils.absolutePath()` now collapses any `file:///…` value to its basename and re-resolves against the current `applicationDataDirectory`, so legacy rows self-heal the next time they're read. Pure decision extracted to `resolvePhotoLocation()` and unit-tested.
- **Write relative:** `taxa.setPhoto` / `sample.setSitePhoto` persist the relative name, not the absolute `nativePath`.
- **Display resolve:** `PhotoSelect.setImage` resolves the stored path before assigning to `ImageView.image` (the SiteDetails display path passed it raw).
- Updated [docs/patterns/photo-paths.md](docs/patterns/photo-paths.md).

Addresses [Trello WB-88](https://trello.com/c/2RZVGoQq/88-inconsistent-sync) — and the underlying real-device bug behind WB-89 (the merged #256 guardrail test couldn't reproduce it on the sim because the container never changes mid-session).

**Evidence:** an on-device diagnostics DB dump showed two distinct container UUIDs across rows while every referenced file was present in the *current* container's Documents — confirming the files exist and only the stored path prefix is stale.

## Test plan
- [x] `npx grunt unit-test-node` — passes (incl. new `test/util/PhotoUtils_spec.js`)
- [ ] CI green (iOS + Android)
- [ ] On-device verification: install over a prior build (or change container), open a synced sample, confirm previously-blank creature/site photos now render

### DIFF
--- a/docs/patterns/photo-paths.md
+++ b/docs/patterns/photo-paths.md
@@ -1,6 +1,15 @@
 # Photo paths
 
-- **User-taken photos** live in `Ti.Filesystem.applicationDataDirectory` and are stored as **relative paths** (no leading `/`).
+- **User-taken & downloaded photos** (site photos, creature/taxon photos) live in `Ti.Filesystem.applicationDataDirectory` and are stored as **relative names** (no leading `/`), e.g. `taxon_12_171_1758403516.jpg`.
 - **Taxonomy reference images** live in `Ti.Filesystem.resourcesDirectory` and are stored as **absolute paths** (leading `/`).
 
-`PhotoUtils.absolutePath()` handles both conventions — call it whenever you need the resolved filesystem path for either kind.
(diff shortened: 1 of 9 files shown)


TheCodeSharman:
[Photos missing on iOS](https://trello.com/c/Fj69NkBA/89-photos-missing-on-ios)

PR #266, opened 2026-05-21
WB-8: Update sync logic to separate the upload of a new sample from the sync of historical samples

TheCodeSharman:
## Summary
- **Why:** a long historical download phase blocked all upload progress, so a freshly-submitted sample couldn't reach the server until thousands of photo downloads drained (see prior WB-12 device verification). WB-8 separates the two concerns.
- Split the single sync path into two operations over one runner: `forceSync()` (the Sync button) downloads then uploads the full history; `uploadPending()` (survey submit / resume) uploads the pending-upload queue only, with no download.
- The app no longer **initiates** a sync on login, app boot, network change, or a 30-minute timer. Survey submit now triggers an upload-only flush of the queue.
- A **user-requested full sync is resilient**: its intent is persisted in a `fullSyncPending` flag (`Ti.App.Properties`), set on Sync and cleared only on success/logout, so it resumes to completion across network drop/restore, backgrounding→foreground, and app restart — but is never auto-*initiated*. Pending uploads need no flag: the upload queue is their persistent state.
- `resumeInterruptedWork()` is the single funnel for app-start, login, network-restore, and foreground — resumes a pending full sync if flagged, else flushes queued uploads, else does nothing. Re-running is safe because `downloadSamples()`/`uploadSamples()` are idempotent.
- Renames for clarity: `SampleSync.forceUpload` → `forceSync`/`uploadPending`; `Survey.forceUpload` → `Survey.uploadNewSample`; `SyncFeedback.start()` → `forceSync()`. Test doubles updated to match.

Implements [Trello WB-8](https://trello.com/c/rU0vkDyg/8-update-sync-logic-to-separate-the-upload-of-a-new-sample-from-the-sync-of-historical-samples) — separating new-sample upload from historical-sample sync, with manual full sync plus resilient resume of an in-progress sync.

## Test plan
- [x] `npx grunt unit-test-node` — passes (173, incl. SyncFeedback forwards to `forceSync()`)
- [x] iOS device specs via LiveView — SampleSync 40 passing (6 new WB-8 orchestration/resilience specs), SampleHistory/Notes/Main 15 passing
- [ ] `npx grunt --platform=android --simulator unit-test` — clean-bundle run
- [ ] Visual/behavioural check on iOS simulator: no auto-sync on login/boot; submit uploads only; Sync runs full download+upload; full sync resumes across airplane-mode toggle, background/foreground, and force-quit relaunch
- [ ] Visual/behavioural check on Android emulator (activity `resume` foreground path)

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -589,10 +589,15 @@ const KobitonAPI = require("./features/support/kobiton");
 
     grunt.registerTask("cucumber", function () {
       const done = this.async();
-      const tags = grunt.option('cucumber-tags') || 'not @skip';
+      const platform = grunt.option('platform');
+      // Default excludes @skip everywhere; on iOS it also excludes @skip-ios
+      // (e.g. the form-login scenario whose iOS "Save Password?" sheet leaks
+      // across scenarios — WB-87). Android still runs those.
+      const tags = grunt.option('cucumber-tags')
+        || (platform === 'ios' ? 'not @skip and not @skip-ios' : 'not @skip');
       const name = grunt.option('grep') || null;
       const appiumOptions = {
-        platform: grunt.option('platform'),
+        platform,
         isSimulator: !!grunt.option('simulator'),
         host: grunt.option('kobiton') ? 'kobiton' : 'local',
       };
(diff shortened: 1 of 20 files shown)


TheCodeSharman (commented):
Code seems fine - lets do some on device tests to smoke test the build before release to test

PR #332, opened 2026-06-18
WB-159: Select photos from the phone gallery

TheCodeSharman:
## Summary
- Adds a gallery-import button to the `PhotoSelect` component so users can attach a photo they took **outside the app** (the core WB-120 "adding a sample after the fact" use case) — not just freshly captured ones, on the bug-ID (EditTaxon) and site-ID (SiteDetails) screens.
- Reuses the existing `PhotoUtils.optimisePhoto` + `savePhoto` relative-path pipeline so gallery imports behave identically to captured photos. `Ti.Media.openPhotoGallery`'s `success` callback returns the same `{ media: <Ti.Blob> }` shape as `showCamera`, so it drops straight into the existing `photoCapturedHandler` → optimise → save path (wrapped behind `lib/ui/Gallery`).
- Permissions: iOS uses `NSPhotoLibraryUsageDescription` (already declared) via a request-then-open gate. Android adds `READ_MEDIA_IMAGES` (API 33+) and opens the system picker directly (`hasPhotoGalleryPermissions` returns false on API 36 even when granted, so the gate is bypassed on Android).
- One end-to-end acceptance scenario covering the full WB-120 flow: existing site + bug photos imported from the gallery, survey **back-dated** via the native date picker, sample stored in history — driving the **real** OS gallery + date pickers on both platforms (no test doubles).

[Trello WB-159](https://trello.com/c/ZyiZeoCR/159-select-photos-from-the-phone-gallery).

## Notes for reviewers
- The acceptance scenario drives the real native pickers: iOS PHPicker + inline calendar; Android system photo picker + Material date dialog (year list scrolled back to the target, months stepped by header comparison).
- Android-specific fix: the gallery picker pauses geolocation, and the emulator's first GPS fix reports `accuracy=100`m (which the app rejects, gate `<100`m), converging only after a few seconds of listening. The scenario waits for the on-screen GPS lock before opening the picker so it doesn't race convergence (was flaking ~2/3 of runs).

## Test plan
- [x] Device unit spec (`PhotoSelect_spec.js`) — gallery button visible when editable, hidden in readonly
- [x] Acceptance scenario: existing photos via gallery + back-dated survey date (WB-120 use case) — green on iOS (3/3) and Android (3× liveview + fresh non-liveview build)
- [x] Android 13+ gallery-permission behaviour confirmed (READ_MEDIA_IMAGES + direct picker open)


features/add_sample_after_the_fact.feature
```diff
@@ -0,0 +1,20 @@
+Feature: Add a sample after the fact
+
+A user who collected a sample in the field — recording the photos on a
+separate camera and the date on paper — can enter it into the app later.
+They choose their existing photos from the phone's gallery and back-date
+the survey to the day they actually collected it.
+
+@gallery
+Scenario: Enter a previously-collected sample from gallery photos with a back-dated survey date
```
TheCodeSharman: Can you memove the odd use of hypens here thanks.

features/add_sample_after_the_fact.feature
```diff
@@ -0,0 +1,20 @@
+Feature: Add a sample after the fact
+
+A user who collected a sample in the field — recording the photos on a
```
TheCodeSharman: Can you rewrite the scenario using "I want to..." for conssitency with other scenario text. Also can you remove the parenthical hypen usage.

features/add_sample_after_the_fact.feature
TheCodeSharman: Done in b0c3c0d4 — scenario title now reads "previously collected … back dated" without the hyphens.

features/add_sample_after_the_fact.feature
TheCodeSharman: Done in b0c3c0d4 — rewritten as "I want to enter a sample I collected in the field earlier…" in first-person style, em-dashes removed.

PR #351, opened 2026-07-20
WB-188: Add menu icon to home screen

TheCodeSharman:
## Summary
- Adds `MenuViewModel` — the home screen's login label, version label/colour and navigation actions were previously imperative controller code, so none of it was reachable from a Node test. It now has 12 Node-speed specs.
- Rewires `controllers/Menu.js` through `bindView()`, dropping the inline `onClick` attributes from `Menu.xml`. The controller keeps only what a port would throw away: the `MethodSelect` Alloy sub-controller and the `Ti.UI.AlertDialog` logout confirmation.
- Adds the **Academy** button, fourth from the left and immediately before About, per the Figma mock. It is deliberately unwired — the card asks for a no-op until the training feature exists.

The refactor lands first and separately from the button, so the behaviour-preserving change is reviewable on its own.

[Trello WB-188](https://trello.com/c/rV2g26YQ/188-add-menu-icon-to-home-screen).

## Notes for review
- **`academy-icon.png` has an export artifact** — a stray 1px vertical line down its right edge, visible on screen just right of the graduation cap. Carried over from the icon commit on the original branch; worth a re-export.
- **Two dead functions left untouched** in `Menu.js`: `mayflyClick()` and `helpClick()`. Neither is referenced from `Menu.xml` and both predate this work. Left in place rather than swept up here — happy to remove in a separate PR.
- The Academy button has no `onClick` binding at all rather than a no-op ViewModel method. That reads as "not wired yet" instead of adding speculative structure, but it does mean nothing marks it as intentional — say the word if you'd rather see an explicit no-op.

## Test plan
- [x] `npx grunt unit-test-node` — 332 passing
- [x] `npx grunt --platform=ios --simulator unit-test` — 393 passing, clean bundle (no LiveView)
- [x] Visual check on iOS simulator — Academy renders 4th, matches the mock
- [x] Visual check on Android emulator
- [x] Acceptance tests


walta-app/app/controllers/Menu.js
```diff
@@ -1,51 +1,59 @@
...
+
+var vm = new MenuViewModel({
+  cerdiApi: Alloy.Globals.CerdiApi,
+  topics: Topics,
+  environment: Alloy.CFG.environment,
+  version: Ti.App.version
+});
+
+bindView($, vm, {
+  appVersion:      { text: "versionLabel", color: "versionColor" },
```
TheCodeSharman, line 16: I notice there is no mention of academy in this bindView config? That seems odd, the ViewModel is supposed to encapsulate every button.

walta-app/app/controllers/Menu.js
```diff
@@ -1,51 +1,60 @@
...
```
TheCodeSharman, line 16: Good catch — that was the inconsistency I flagged in the PR description, and you're right that leaving it out breaks the "ViewModel owns every button" rule.

Fixed in 01f80343: added an inert `academy()` no-op to the ViewModel and bound it via `onClick` like the others. It does nothing yet (per the card — the training feature isn't built), but the button now routes through the ViewModel and there's a clear seam to wire up later. Pinned by a VM spec asserting it fires no topic or event, and the Menu device specs (8) still pass.

PR #364, opened 2026-07-24
WB-189: Add create training session start screen to allow entering session id.

TheCodeSharman:
## Summary
- **New Titanium-free modal-controller pattern.** Modals are now a first-class navigation concept: `Navigation.openModal`/`closeModal` name the concept and delegate the Titanium glue (build the Alloy overlay `$`, overlay it, tear it down) to `View`. A `lib/controllers/<Name>` screen controller — plain JS, no `Ti.*` — owns the non-Titanium wiring (build the ViewModel, `bindView`, route events to injected orchestration). `lib/controllers/registry.js` catalogues which modals use the pattern; absent ones still open as a plain overlay, so legacy modals migrate incrementally.
- **Academy modal (the feature).** Tapping Academy on the menu opens the modal (title, intro, three digit-code boxes, Close/Start, ✕). `AcademyViewModel` holds the 3-digit code + `startEnabled` gating; Start is an inert seam for now.
- **Two-way `bindView` binding** (`twoWay()`): typed text flows widget→VM through the bind map (the digit boxes), removing hand-wired `onChange`. Reuses the existing diff-guard to close the feedback loop. Absorbs the previously-unbuilt WB-160 north-star step.
- **Acceptance scenario** written to *extend* into the full training-session flow later, not be replaced.
- Docs: new `docs/patterns/modals.md`; `twoWay` documented in `viewmodels.md`.

Delivers [Trello WB-189](https://trello.com/c/oEA0Gl1j/189-add-create-training-session-start-screen-to-allow-entering-session-id). Deferred (separate cards): migrating the `MethodSelect` identify overlay onto this pattern, the training-session flow Start eventually launches, and deeplink-straight-to-modal.

Known nit for review: in landscape with the numeric keyboard up, the keyboard overlaps the lower half of Close/Start (functionally fine — entry dismisses it before Start). Polish candidate.

## Test plan
- [x] `npx grunt unit-test-node` — passes (383): two-way binding, AcademyViewModel, Academy controller, Navigation modal API
- [x] iOS acceptance (LiveView): `Enter a training session code` — 1 scenario / 8 steps green
- [x] iOS visual check — modal matches the UX mockup; two-way binding shows 1·2·3 in the boxes
- [ ] Android acceptance + clean (non-LiveView) build — via CI


walta-app/app/lib/controllers/Academy.js
```diff
@@ -0,0 +1,35 @@
+const AcademyViewModel = require("viewmodels/Academy");
+const bindView = require("util/bindView");
+const { twoWay } = bindView;
+
+// Titanium-free screen controller for the Academy modal. View owns the Titanium
+// widgets (the $ bag) and hands them in as `view`; this file owns everything
```
TheCodeSharman: i feel liike some of this comment is too much detail, lets keep the design discussion inside the docs.

TheCodeSharman: Agreed — trimmed to a short pointer, and did the same for the sibling comments in `registry.js` and `View.openModal` that repeated the same narrative. The design write-up lives in `docs/patterns/modals.md`. (931a418)

walta-app/app/lib/controllers/Academy.js
```diff
@@ -0,0 +1,46 @@
...
+  digit3:       { title: "digit3" },
+  digitPicker:  { visible: "pickerVisible", onClick: "cancelPicker" },  // tap backdrop to dismiss
+  startButton:  { enabled: "startEnabled", onClick: "start" },
+  closeButton:  { onClose: "close" },   // the ✕ (CloseButton Require)
+  cancelButton: { onClick: "close" },   // the "Close" text button
+};
+
+// A box tap carries which box (0-2), a key tap carries which digit — arguments
+// bindView's arg-less event handlers can't pass, so wire them here.
+function wireDigitPicker(view, vm) {
```
TheCodeSharman: Ok, can we extend bindView() in a way to support this kind of wiring up: this brings back Titanium code into this controller which is supposed to be Titanium Free. *ALL* titanium specific wiring needs to be abstracted behind the `bindView` interface.

TheCodeSharman: Agreed — this is a real leak: `wireDigitPicker` calls `addEventListener`/`removeEventListener` directly, so raw Titanium is back in a controller that's meant to be Ti-free. It's only here because these events carry an argument (which box, which digit) and bindView's event bindings are currently arg-less (`onClick: "method"` → `vm.method()`).

Fix: extend bindView with an argument-carrying event marker, mirroring the existing `twoWay(prop)` marker — `call(method, ...args)`:

```js
const { call } = bindView;
const BINDINGS = {
  digit1: { title: "digit1", onClick: call("startEditing", 0) },
  digit2: { title: "digit2", onClick: call("startEditing", 1) },
  digit3: { title: "digit3", onClick: call("startEditing", 2) },
  digitPicker: { visible: "pickerVisible", onClick: "cancelPicker" },
  startButton: { enabled: "startEnabled", onClick: "start" },
  closeButton: { onClose: "close" },
  cancelButton: { onClick: "close" },
};
for (let d = 0; d <= 9; d++) BINDINGS["keypad" + d] = { onClick: call("pickDigit", d) };
```

bindView wires `() => vm.startEditing(0)` etc. behind its own `attachEvent`, so `wireDigitPicker` and both `addEventListener`/`removeEventListener` calls are deleted — the controller goes back to declarative BINDINGS + create/dispose only, with zero `Ti.*` surface. Plain string handlers keep working unchanged, and teardown stays inside bindView's existing unbind.

I'll do it test-first in `bindView_spec` (arg-carrying handler fires with the right args + is torn down on unbind), then refactor the controller. One open question: happy with `call` as the marker name, or prefer something like `invoke`/`handler`?

TheCodeSharman: Done — went with `call` (easy to rename if you'd rather `invoke`/`handler`). bindView now has a `call(method, ...args)` marker alongside `twoWay(prop)` (ceabd47), and the controller's `wireDigitPicker` is gone — box/keypad taps are declarative `onClick: call("startEditing", i)` / `call("pickDigit", d)` bindings, teardown lives in bindView's `unbind`, and the controller has zero `Ti.*` surface (44d4919). Documented the marker in `docs/patterns/viewmodels.md` (ab7b584).

Covered test-first in `bindView_spec` (arg-carrying handler fires with the right args, is torn down on unbind, validates the method exists); full Node suite green (397), and the Academy acceptance scenario still passes end-to-end on the iOS sim.

TheCodeSharman: Excellent - happy with `call`

walta-app/app/views/Academy.xml
```diff
@@ -0,0 +1,43 @@
+<Alloy>
+	<View class="overlay">
+		<View class="window academyWindow" layout="vertical">
+			<View class="titlebar">
+				<Label id="academyTitle" class="titleText" accessibilityLabel="Academy">Academy</Label>
+				<Require id="closeButton" src="CloseButton"/>
+			</View>
+			<Label class="academyIntro">Test your macroinvertebrate identification chops: navigate to https://waterbugblitz.org/academy (on a separate device) and following along with the video tutorial.</Label>
```
TheCodeSharman:
```suggestion
			<Label class="academyIntro">Test your macroinvertebrate identification chops: navigate to https://waterbugblitz.org/academy (on a separate device) and follow along with the video tutorial.</Label>
```

PR #372, opened 2026-07-29
chore(gps): remove stray WB204DIAG log line

TheCodeSharman:
Removes an instrumentation `Ti.API.info` that slipped into the merged GPS fix (b4e2c3fc) via a git-add slip during a reset+recommit. Harmless log line, but doesn't belong in production.


TheCodeSharman:
Superseded — WB-208 (the SiteDetails GPSLOCK leak fix, merged in #374) removed this WB204DIAG log line as part of refactoring that handler. The line is no longer on main, so this PR is a no-op. Closing.

PR #375, opened 2026-07-29
WB-206: guard postlayout .size reads (getWindow null-activity race)

TheCodeSharman:
## Summary
- **Stacked on WB-208 (#374)** — base is the leak-fix branch so this diff is only the guard. CI here validates the **combined** fix (leak fix + guard), i.e. the real target state of `main`.
- The residual ~1/3 Android-acceptance flake after WB-208 is an **independently-triggerable** window-transition race: opening a screen mid-transition, Android fires `postlayout` while the current activity is transiently null; handlers reading view geometry then throw `getWindow()` on null and abort mid-layout — `AnchorBar.updateTitleWidth` (`$.rightTools.size.width`) and `SampleTray` tray-init (`$.content.size.height`) — leaving the header/tray (and the `Add Sample` button) unrendered → step times out.
- **Fix:** `util/TiHacks.attemptLayout(fn)` runs the layout work, swallows **only** that transient `getWindow()` failure (returns false so the caller retries on the next `postlayout`), and **rethrows anything else**. AnchorBar re-fires naturally; SampleTray keeps its one-shot listener until layout completes.

[Trello WB-206](https://trello.com/c/urNDbIly). Part of the WB-207 cluster.

## Caveat (please read before merging)
This is a **guard for a Ti platform timing race**, verified deterministically only at the unit level. The acceptance-level effect is **intermittent and CI-only** — needs several green Android-acceptance runs to trust. I've kept it low-risk (no re-timing of the window close, which regressed in an earlier attempt; retry-on-next-postlayout only).

## Test plan
- [x] `npx grunt unit-test-node` — 369 passing incl. new `attemptLayout` spec (RED→GREEN)
- [x] `npx grunt --platform=ios --simulator --grep="SampleTray|AnchorBar" unit-test` — 35 passing (no regression)
- [ ] Android acceptance green across **several** CI runs (the real proof)

### DIFF
--- a/walta-app/app/controllers/SampleTray.js
+++ b/walta-app/app/controllers/SampleTray.js
@@ -525,17 +525,27 @@ $.TopLevelWindow.addEventListener("dragend", handleDragEnd );
 */
 
 $.content.addEventListener( "postlayout", function initEvent() {
-  // Guard the size-dependent tray layout: a postlayout that fires mid
-  // window-transition (transiently-null activity) throws on the .size read, which
-  // used to leave the tray — and its Add Sample button — unrendered. Keep the
-  // listener so the next postlayout retries; only remove it once layout completes.
-  if ( attemptLayout(function() {
+  $.content.removeEventListener( "postlayout", initEvent );
+  initTrayWhenAttached();
+});
+
+// content's 'postlayout' can fire mid window-transition, when the current
+// activity is transiently null and reading .size throws getWindow()-on-null.
+// attemptLayout swallows that, but content's postlayout is one-shot and won't
+// re-fire — so poll briefly until the activity is attached and the size-dependent
+// tray layout (and its Add Sample button) actually renders. initializeTray reads
+// a size on its first line before creating anything, so a swallowed attempt is a
+// no-op and retrying is safe.
+function initTrayWhenAttached( attempt ) {
+  attempt = attempt || 0;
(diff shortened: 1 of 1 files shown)


TheCodeSharman:
**CI finding (2 Android-acceptance samples): the guard is necessary but not sufficient.**

Both samples: **getWindow crash count = 0** (the guard eliminates the V8 crash — down from 2), but `~Add Sample.` still times out. So the guard swaps the crash for a **silent non-render**: `SampleTray`'s one-shot `content` `postlayout` fires once mid-transition, `attemptLayout` skips it to retry — but `content` `postlayout` doesn't re-fire, so `initializeTray()` never completes and the tray/Add-Sample button never mount.

**Real fix direction** (for whoever picks this up): keep `attemptLayout` (it correctly detects the transient), but give the retry a **reliable trigger** instead of relying on `content` re-firing `postlayout` — e.g. re-attempt tray init on the TopLevelWindow's settled signal (`window-opened` / post-safe-area) or a short `setTimeout` re-try, so init actually runs once the activity is attached. `AnchorBar.updateTitleWidth` is fine as-is (its `postlayout` re-fires naturally).

Leaving this draft for review — the crash-guard is a sound building block; the tray-init retrigger is the missing piece. WB-208 (the leak fix this is stacked on) is the independently-mergeable win.

PR #378, opened 2026-07-30
WB-191: SampleTray ViewModel rearchitecture (scroll windowing)

TheCodeSharman:
## Summary
- **Spike 0 landed** — answers the card's highest-risk question (how much of the fluid virtualized scroll can move off Titanium) with on-device proof before the full port.
- New Titanium-free `SampleTrayViewModel` owns the tile/index arithmetic, the layout geometry (endcap/middle/tray widths, tile offsets — all derived from the measured viewport in dip), and the scroll windowing (`roundToTile` → `visibleRange` → `syncWindow`). Node-tested **bit-for-bit** against the old controller math, including the `i>=0` clamp and the `rightEdge-1` off-by-one.
- `syncWindow` is a **pure per-scroll query** returning a `{toAdd, toUpdate, toRelease}` delta (add/update entries carry their computed `left`/`width`/`collectionIndices`) — deliberately **not** routed through `notifyListeners`, so per-frame cost stays one call + a small delta.
- `controllers/SampleTray.js` now drives the scroll handler off `vm.syncWindow(offset)` and applies the delta with the existing Ti view create/update/remove; `roundToTile`/`addTiles`/`releaseTiles` drop out of the controller.

Refactor-only, **zero behaviour change** — the first slice of [Trello WB-191](https://trello.com/c/4HoxjrCl/191-add-the-feedback-screen). Establishes the VM boundary; the `SampleTaxaIcon` `bind()` port, hole descriptors, `abundanceVisible` tidy, training mode + tick/cross feedback all land in later stacked PRs.

## Test plan
- [x] `npx grunt unit-test-node` — passes (14 VM tests, bit-for-bit geometry + windowing)
- [x] On-device `SampleTray` spec unchanged on iOS (32/32), incl. `scrolling a long tray` virtualize-under-scroll
- [x] Fluidity fling confirmed on real iPhone 17e **and** physical Android vs current build — both fluid (the spike gate)
- [ ] On-device `SampleTray` spec on Android emulator
- [ ] `SampleTray` visual-regression baseline stays 0px

### DIFF
--- a/test/viewmodels/SampleTray_spec.js
+++ b/test/viewmodels/SampleTray_spec.js
@@ -147,6 +147,18 @@ describe("SampleTrayViewModel", function () {
       // scrollx 0 gives leftEdge -1; the window must start at tile 0, not -1.
       expect(vm.syncWindow(0).toAdd.map(t => t.tileNum)).to.deep.equal([0, 1, 2, 3]);
     });
+
+    // The controller clears every cached tile view when it rebuilds the tray;
+    // resetWindow drops the materialized bookkeeping so the next syncWindow
+    // re-adds the whole window from scratch.
+    it("re-adds the whole window after resetWindow", function () {
+      const vm = vmWithViewport(30);
+      vm.syncWindow(0);
+      vm.resetWindow();
+      const { toAdd, toUpdate } = vm.syncWindow(0);
+      expect(toAdd.map(t => t.tileNum)).to.deep.equal([0, 1, 2, 3]);
+      expect(toUpdate).to.deep.equal([]);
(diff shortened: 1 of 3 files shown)


TheCodeSharman:
Superseded by #385, which ports SampleTray onto the full presentation-DSL architecture (injected bindView + the collection binding + the registry) that landed on main via #382/#383 after this branch was cut. The VM spike's geometry/windowing arithmetic carries over into #385's SampleTrayViewModel (syncWindow/_materialized folded into a visibleTiles getter the binder's keyed diff drives).

PR #381, opened 2026-07-31
WB-213: View screen-controller seam — tested, drives windows too, LiveView-unblocked

TheCodeSharman:
## Summary
`View` (`lib/logic/View.js`) is the load-bearing seam of the north-star — the only code (with `bindView`) that touches Alloy controllers — and it had **zero specs**. This branch covers it, makes it testable, and generalizes it so a *window* gets the same Titanium-free screen-controller treatment a modal already had.

- **Specs** — new device spec `walta-app/app/spec/View_spec.js` characterizing the seam: `openView` opens a window + tracks `currentController` + builds/disposes the window's lib controller; `openModal` overlays a modal + clears it on close; the lib controller is built with `{ view, close, services, palette }` from both the **injected** and the **default** registry.
- **Injectable registry** — the screen-controller registry is injectable (`services.screenControllers`), defaulting to `require("mvvm/controllers/registry")`, so the seam is testable with a fake controller.
- **`openView` generalized** — a window now instantiates its `lib/mvvm/controllers/<name>` (if registered) via `attachScreenController` and disposes it on the window's `close` event — the same registry path `openModal` uses. `lib/controllers/` moved to `lib/mvvm/controllers/` to end the namespace clash with Alloy's own `controllers/`.
- **Docs** — de-modal-ified: new `docs/patterns/screen-controllers.md` owns the tier (factory contract, three MVVMC tiers, port story); `modals.md` trimmed to modal-specific glue; `screen-plumbing.md` + `viewmodels.md` corrected.

First slices of [Trello WB-213](https://trello.com/c/HKhcQDxG).

## LiveView dev-loop fix (was: ⚠️ finding to review)
The default `require("mvvm/controllers/registry")` chain mis-resolved **under LiveView**: the fork's Alloy component plugin matched `/controllers/` anywhere in a request and redirected the lib require to the Alloy shell (and fed it to the Alloy component compiler). Fixed in the LiveView fork by anchoring that regex to `appDir`/`alloy` — mirroring the earlier model-plugin fix, but also keeping the un-aliased virtual `/alloy/controllers/*` request matching (a pure `appDir` anchor would break every app controller). The pinned `liveview` commit is bumped here. The re-added `View_spec.js` default-registry test now passes under **both LiveView and a real bundle build** (6/6). Real (non-LiveView) builds — CI included — were never affected.

## Test plan
- [x] `View` seam device spec on iOS — LiveView + real build, 6 passing
- [x] `View` seam device spec on Android — Android unit-test CI job green (real build)
- [x] `openView` generalized to drive `lib/mvvm/controllers/<name>` for windows
- [x] De-modal-ify the registry naming + `docs/patterns/`
- [x] Verify registry-require reliability in a real (non-LiveView) build

### DIFF
--- a/docs/patterns/modals.md
+++ b/docs/patterns/modals.md
@@ -1,8 +1,9 @@
 # Modal screens
 
 How a modal (an overlay above the current window, e.g. the Academy training-session
-screen) is opened, wired, and torn down — as a first-class navigation concept with
-the Titanium-specific glue kept in one place.
+screen) is opened and torn down. A modal shares the Titanium-free screen-controller
+split with windows — see [screen-controllers.md](screen-controllers.md); this page
+covers only the modal-specific overlay glue.
 
 ## Opening a modal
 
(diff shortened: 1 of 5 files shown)


docs/patterns/modals.md
```diff
@@ -21,7 +21,7 @@ work to `View.openModal` (`lib/logic/View.js`), which:
 
 1. `Alloy.createController(name, args)` — builds the overlay's widgets (`$`).
 2. adds the overlay view onto the current window.
-3. looks the modal up in `lib/controllers/registry.js` and, if present, hands its
+3. looks the modal up in `lib/mvvm/controllers/registry.js` and, if present, hands its
    Titanium-free screen controller the widgets plus a `close` callback.
```
TheCodeSharman: Is this wrong - the mvvm/controllers/registry isn't jsut for modal its for *any* view

PR #382, opened 2026-07-31
WB-214: Port Menu to the View screen-controller seam

TheCodeSharman:
## Summary
Ports the Menu screen onto the `View` screen-controller seam, then (per review) empties the Alloy controller so **all** business logic lives in the Titanium-free mvvm layer. This is the clean precedent for future screen ports.

Menu's Alloy controller is now just:
```js
exports.baseController = "TopLevelWindow";
$.name = "home";
```

**How the logic moved out:**
- **`lib/mvvm/controllers/Menu`** (Ti-free) builds the VM from the services bag, `bindView`s it, and confirms logout via the `Dialogs` seam. Button clicks → VM → Topics as before.
- **MethodSelect → a real mvvm modal.** New `Topics.SELECT_METHOD` (routed by `Main` → `Navigation.openModal`) carries the caller's payload; new Ti-free `mvvm/controllers/MethodSelect` routes the shell's selection events to the nav Topics (KEYSEARCH/SPEEDBUG/BROWSE/IDENTIFY). Menu's `identify()` and **SampleTray's `startIdentification`** both just fire `SELECT_METHOD` — SampleTray sheds ~35 lines of inline overlay. The screen-controller factory now also receives the open `args`.
- **Native logout dialog → `Dialogs` seam** (`lib/logic/Dialogs.js`, `confirm() → Promise<bool>`), injected via services, fakeable in tests.
- **Menu's close handler dropped** — the `TopLevelWindow` base already does `$.destroy()/$.off()` on close; its only unique teardown was the overlay.
- **Dead code removed** (`mayflyClick`/`helpClick`/`mayflyMap`).
- **`makeTestServices()` fixture** gives app-boot harnesses (Main/Notes/Menu) the full services bag instead of hand-rolled partials.

Review comments all addressed inline. [Trello WB-214](https://trello.com/c/XtVzN7zD).

## Note on test coverage
The MethodSelect selection→topic routing tests moved from the callers (Menu/SampleTray device specs) into `test/controllers/MethodSelect_spec.js` (Node) where that logic now lives — Menu/SampleTray specs assert they fire `SELECT_METHOD`. Net coverage preserved, at a cheaper layer.

## Follow-up
True *zero*-controller (view-only, no `controllers/Menu.js`) needs `baseController = "TopLevelWindow"` — an Alloy mechanism with no XML equivalent — moved into the `View` seam. That touches all ~20 windows, so it's carded separately as [WB-215](https://trello.com/c/feTHxZE6).

## Test plan
- [x] `npx grunt unit-test-node` — 441 passing (Menu, MethodSelect, MenuViewModel lib/VM specs)
- [x] Device specs via LiveView (iOS) — Menu, MethodSelect, View seam, Main, Notes (30), SampleTray (29)
- [ ] iOS + Android unit tests (real build) — via CI
- [ ] Acceptance suite (Menu = entry screen; identify/add-to-sample flows) — via CI


walta-app/app/controllers/Menu.js
```diff
@@ -1,41 +1,24 @@
...
-
-vm.on("identify", openSelectMethod);
-vm.on("confirmLogout", confirmLogout);
+// Residual-Titanium shell. The Menu screen controller
+// (lib/mvvm/controllers/Menu, instantiated by View.openView) builds the
+// view-model and binds it; it routes the identify / confirm-logout intents to
+// the two Ti actions exposed here — the MethodSelect overlay and the logout
+// alert. See docs/patterns/screen-controllers.md.
+$.openSelectMethod = openSelectMethod;
+$.confirmLogout = confirmLogout;
```
TheCodeSharman: These both seem like business logic - how do we rbing them into the framwork agnostics mvvm code?

walta-app/app/controllers/Menu.js
```diff
@@ -1,41 +1,24 @@
...
 
 $.TopLevelWindow.addEventListener('close', function cleanUp() {
-  vm.dispose();
   $.destroy();
   $.off();
   $.TopLevelWindow.removeEventListener('close', cleanUp );
   closeSelectMethod();
 });
```
TheCodeSharman: This is interesting it seems rather generic - possibly hinting at a refactor, thoughts? Maybe not for now.

walta-app/app/controllers/Menu.js
```diff
@@ -1,41 +1,24 @@
...
 
-function confirmLogout() {
+function confirmLogout(onConfirm) {
   var dialog = Ti.UI.createAlertDialog({
```
TheCodeSharman: Again seems like this is businmess logic and needs to be incorporated into the mvvm codebase? Suggestions welcome: perhaps we could have a fake controller to represent the TItanium native dialogs?

walta-app/app/controllers/Menu.js
```diff
@@ -44,27 +27,12 @@ function confirmLogout() {
...
   dialog.addEventListener('click', function(e) {
     if (e.index === 0) {
-      vm.logOut();
+      onConfirm();
     }
   });
   dialog.show();
 }
 
-function mayflyClick() {
```
TheCodeSharman, line 53: Excellent - we need to clean up any dead code. Thanks.

walta-app/app/controllers/Menu.js
```diff
@@ -44,27 +27,12 @@ function confirmLogout() {
...
-      $.mayflyMap.cleanUp();
-      $.mayflyMap = null;
-  });
-}
-
-function helpClick() {
-  Topics.fireTopicEvent( Topics.HELP, null );
-}
-
 function closeSelectMethod() {
```
TheCodeSharman: So these methods are about business logic, the plumbing to opening Topics.* on menu clicks and the MethodSelect modal, these aren't Titanium specific, they are business logic. So clearly we need to make selectmethod a mvvm modal. The othrs are plumbning that should be in the mvvm controller.

TheCodeSharman:
As a general comment, for this to be complete, I would expect Menu to not need *any* alloy controller code at all, currently there is a bunch of business logic that should be in the mvvm controller.

walta-app/app/controllers/Menu.js
TheCodeSharman: Both gone. `identify` now flows VM → `Topics.SELECT_METHOD` → `Navigation.openModal("MethodSelect")` (MethodSelect is a real mvvm modal). Logout confirm moved behind the `Dialogs` seam, driven by the mvvm controller. Menu's controller is now just `baseController` + `$.name`.

walta-app/app/controllers/Menu.js
TheCodeSharman: You were right — it was fully redundant. The `TopLevelWindow` base controller already does `$.destroy()/$.off()` on close; the only unique bit was tearing down the overlay, which is gone now that MethodSelect is a modal. Dropped the whole handler.

walta-app/app/controllers/Menu.js
TheCodeSharman: Done — added a `Dialogs` seam (`lib/logic/Dialogs.js`, `confirm() → Promise<bool>`), injected via services and fakeable in tests. The mvvm controller does `if (await services.dialogs.confirm(...)) vm.logOut();` — Ti-free, VM owns the decision. (`View.askDiscardEdits` can migrate here later.)

walta-app/app/controllers/Menu.js
TheCodeSharman: Removed (`mayflyClick`/`helpClick`/`mayflyMap` were dead — only the separate MayflyEmergenceMap screen + a commented-out route referenced mayfly).

walta-app/app/controllers/Menu.js
TheCodeSharman: Agreed — MethodSelect is now an mvvm modal. The selection→topic routing lives in Ti-free `mvvm/controllers/MethodSelect`; Menu and SampleTray just fire `SELECT_METHOD` with their payload (`allowAddToSample`/`surveyType`/`unknownBug`). The Alloy shell keeps only the dynamic entry-building (residual Ti).

PR #385, opened 2026-08-03
WB-191: Port SampleTray onto the MVVM presentation DSL

TheCodeSharman:
## Summary
- **Supersedes #378.** #378 carried only the earlier VM spike, built before the presentation-DSL architecture (injected `bindView`, the `collection()` binding, the registry, the source-in-services-bag) landed on `main` via #382/#383. This ports the whole screen onto that architecture instead.
- **Ti-free `lib/mvvm/controllers/SampleTray`** declares the tray as two convention collections on `$.tray` — the fixed endcap and the windowed interior tiles — plus the tray width. Mirrors `SampleHistory`.
- **No escape hatch.** The scroll listener is Titanium and stays in the shell: it feeds `vm.setScrollOffset(dip)` (the offset read + DIP conversion stay Ti-side), which notifies so the binder's keyed diff reconciles the visible window. The explicit-adapter `scrollEvent`/`onScroll` form is now unused across the whole app (removable in a follow-up tidy).
- **All geometry, windowing and hole-kind logic lives in the Node-tested `SampleTrayViewModel`** (a two-level `ChangeNotifier` tree: tray → tile/endcap → icon VMs). `SampleTraySource` holds the model reads behind the source seam.
- **`SampleTrayTile` / `SampleTrayEndcap`** are first-class components (mvvm binds geometry; the Alloy controller renders holes via `TrayHoleRenderer`). **`SampleTaxaIcon`** becomes `exports.bind(vm)` and owns its own edit tap.
- **The Alloy `SampleTray` shell drops to residual Titanium** — viewport measurement, scroll offsets, `scrollToRightEdge`, the `EditTaxon` overlay.

Refactor-only, **zero behaviour change** — the first slice ("PR 1") of [Trello WB-191](https://trello.com/c/4HoxjrCl/191-add-the-feedback-screen). The `SampleTaxaIcon` tick/cross overlay (PR 2, commits the staged icon assets) and the training-mode + Assess feature (PR 3) land in later stacked PRs; the `icons.svg` + tick/cross PNGs are intentionally left uncommitted here.

The on-device spec now opens through the View seam (like `SampleHistory`) so the mvvm controller wires the bindings; the test bodies are otherwise unchanged, and the two `getTileIndex()` reach-ins fire on the rendered cells the spec already navigates (the internal tile-index datastructure is gone with the port).

## Test plan
- [x] `npx grunt unit-test-node` — passes (SampleTrayViewModel geometry/windowing/holes/notification; 498 total)
- [x] On-device `SampleTray` spec on **iOS — 29/29**
- [x] On-device `SampleTray` spec on **Android — 29/29** (plus-cell tap moved onto the button — a synthetic `fireEvent` doesn't bubble to a container handler on Android)
- [x] `SampleTray` visual-regression baseline **0px on Android** (both CI renderers); iOS advisory pending in CI
- [x] iOS acceptance — `Creature photos remain visible after sync` passes locally after the tap fix (see below)
- [x] Scroll fluidity fling on a real device vs the current build — windowing math unchanged from the spike (already flung fluid on real iPhone + Android); worth a final human confirmation

**Note:** the initial CI run surfaced a real iOS-only regression — a tap whose hit-test target is a Ti view with no click listener raises no bubbling click, so tapping a creature didn't open EditTaxon. Fixed by having `SampleTaxaIcon` own its tap on the silhouette ImageView + the cell root. The device unit spec missed it (synthetic `fireEvent`); acceptance caught it.

### DIFF
--- a/walta-app/app/spec/SampleTray_spec.js
+++ b/walta-app/app/spec/SampleTray_spec.js
@@ -133,6 +133,15 @@ describe( 'SampleTray controller', function() {
     return sorted[i];
   }
 
+  // Poll the rendered tray until the content assertions hold. Settling the scroll
+  // only guarantees the offset; under load the windowed tile a scroll reveals is
+  // created — and its slot images bound — a beat later, so a one-shot assertion
+  // races the render. waitFor swallows the assertion throw and retries to the
+  // deadline, so a genuine mismatch still fails (as a timeout).
+  function assertEventually( assertFn ) {
+    return waitFor( function() { assertFn(); return true; } );
+  }
+
   // A cell is one of two polymorphic slot components: a SampleTaxaIcon (taxon/blank,
   // children = [ padIcon, tapSurface ]; padIcon holds [ icon, abundance ]) or a
   // SampleTrayPlus (add, children = [ plus, tapSurface ]). In both, the tap surface
@@ -433,14 +442,14 @@ describe( 'SampleTray controller', function() {
           .then( scrollSampleTray(0) )           // scroll back to left
           .then( () => trayVm().trayWidth - trayVm().viewWidth )
           .then( (width) => scrollSampleTray(width)() )
-          .then( function() {
+          .then( () => assertEventually( function() {
             var tiles = SampleTray.tray.children;
             var tile = findRightMost( tiles );
             assertTaxaBackground( tile, "images/tiling_interior_320.png" );
             var sampleTaxa = getTaxaIcons( tile );
             expect( sampleTaxa ).to.have.lengthOf(4);
             assertPlus( sampleTaxa[0] );
-          });
+          }));
 
     });
 
@@ -450,36 +459,36 @@ describe( 'SampleTray controller', function() {
           .then( () => trayVm().trayWidth - trayVm().viewWidth )
           .then( (maxX) => scrollSampleTray(maxX)() )
           .then( scrollSampleTray(0) )
-          .then( function() {
+          .then( () => assertEventually( function() {
             var tiles = SampleTray.tray.children;
             expect( tiles.length ).to.be.at.least(4);
 
             tiles.shift(); // discard end cap since that is always static
-          
+
             // assert left most tile
             var tile = findLeftMost( tiles  );
             assertTaxaBackground( tile, "images/tiling_interior_320.png" );
-            
+
             var sampleTaxa = getTaxaIcons( tile );
             expect( sampleTaxa ).to.have.lengthOf(4);
             assertSample( sampleTaxa[0], "/anisops_b.png", "3-5" );
             assertSample( sampleTaxa[1], "/atalophlebia_b.png", "1-2" );
             assertSample( sampleTaxa[2], "/anostraca_b.png", "1-2" );
             assertSample( sampleTaxa[3], "/aeshnidae_telephleb_b.png", "6-10" );
-          });
+          }));
     });
 
     it('should scroll to the far right upon opening', function() {
       return Promise.resolve()
           .then( openSampleTray )
-          .then( () => {
+          .then( () => assertEventually( () => {
             var tiles = SampleTray.tray.children;
             tiles.shift();
             var tile = findRightMost( tiles );
             var sampleTaxa = getTaxaIcons( tile );
             expect( sampleTaxa ).to.have.lengthOf(4);
             assertPlus( sampleTaxa[0] );
-          })
+          }))
     });
 
    
@@ -533,7 +542,7 @@ describe( 'SampleTray controller', function() {
             Alloy.Collections["taxa"].add( Alloy.createModel( "taxa", { taxonId: "1", abundance: "3-5" } ) );
             return scrollDone;
            })
-          .then( () => {
+          .then( () => assertEventually( () => {
             var tiles = SampleTray.tray.children;
             tiles.shift();
             var tile = findRightMost( tiles );
@@ -541,7 +550,7 @@ describe( 'SampleTray controller', function() {
             expect( sampleTaxa ).to.have.lengthOf(4);
             assertPlus( sampleTaxa[0] );
             expect( SampleTray.tray.size.width ).to.be.above( SampleTray.content.size.width );
-          });
+          }));
     });
 
     it('should scroll to the far right after adding 27th taxon', function() {
@@ -590,7 +599,7 @@ describe( 'SampleTray controller', function() {
             Alloy.Collections["taxa"].add( Alloy.createModel( "taxa", { taxonId: "1", abundance: "3-5" } ) );
             return scrollDone;
            })
-          .then( () => {
+          .then( () => assertEventually( () => {
             var tiles = SampleTray.tray.children;
             tiles.shift();
             var tile = findRightMost( tiles );
@@ -598,7 +607,7 @@ describe( 'SampleTray controller', function() {
             expect( sampleTaxa ).to.have.lengthOf(4);
             assertPlus( sampleTaxa[2] );
             expect( SampleTray.tray.size.width ).to.be.above( SampleTray.content.size.width );
-          });
+          }));
     });
 
     // The plus icon opens the identification-method chooser for adding to the


walta-app/app/controllers/SampleTaxaIcon.js
```diff
@@ -1,69 +1,52 @@
...
+  vm = iconVm;
+  readOnlyMode = !!(opts && opts.readonly);
+  var unbind = bindView($, iconVm, {
+    icon:          { image: "image" },
+    abundance:     { text: "abundanceText", visible: "abundanceVisible" },
+    SampleTaxaIcon: { accessibilityLabel: "accessibilityLabel" },
+  });
+  // On iOS a tap whose hit-test target is a Ti view with no click listener does
+  // not raise a click that bubbles to ancestors — so the silhouette ImageView
+  // (what a tap actually lands on) needs the handler directly. The cell root
```
TheCodeSharman: this all should be an a mvvm controller becasue its not titanium specific

walta-app/app/lib/logic/TrayHoleRenderer.js
```diff
@@ -0,0 +1,113 @@
...
+// (its bound iconVm updates the icon in place — no flicker, positional reuse).
+// The VM owns which kind each hole is; this only realises it in Ti.
+
+function createIconContainer(holeWidthCss) {
+  return Ti.UI.createView({ top: 0, width: holeWidthCss, height: "50%" });
+}
+
+function createAddButton() {
+  return Ti.UI.createButton({
+    top: "40%",
```
TheCodeSharman: This is Titnaium specific violating several design rules;firstlyu int view logic not a in a view, secondly its titnaium specific which is only allowed in bindView or in *extreme* cases the residual alloy controller (avoised unless its truely impossible).

walta-app/app/lib/logic/TrayHoleRenderer.js
```diff
@@ -0,0 +1,113 @@
...
+
+// The plus / add-behind cells open the identification-method chooser to add to
+// the current sample. surveyType comes from the live sample model (Ti).
+function fireSelectMethod(e) {
+  var surveyType = parseInt(Alloy.Models.sample.get("surveyType"));
+  Topics.fireTopicEvent(Topics.SELECT_METHOD, { allowAddToSample: true, surveyType: surveyType, unknownBug: true });
+  if (e) e.cancelBubble = true;
+}
+
+function clearSlot(slot) {
```
TheCodeSharman: Suggest making this a first class mvvm controller couled first titnaioum view,.

walta-app/app/lib/mvvm/controllers/SampleTray.js
```diff
@@ -0,0 +1,33 @@
...
+    tray: {
+      width: "trayWidthCss",
+      endcap: collection("endcapTiles", "SampleTrayEndcap"),
+      tiles: collection("visibleTiles", "SampleTrayTile"),
+    },
+  });
+
+  // Hand the VM to the shell so its Titanium listeners (postlayout, scroll, taxa
+  // add/change/remove, scrollToRightEdge) drive the pure windowing.
+  view.attachViewModel(vm);
```
TheCodeSharman: what is attachViewModel why does it exists - why does bindView do this?

walta-app/app/lib/mvvm/controllers/trayCell.js
```diff
@@ -0,0 +1,27 @@
...
+// layout (in TSS). See docs/patterns/screen-controllers.md.
+module.exports = function bindTrayCell({ view, args, bindView }) {
+  const cellVm = args.rowVm;
+
+  const unbind = bindView(view, cellVm, {
+    cell:       { left: "leftCss", width: "widthCss", height: "heightCss" },
+    background: { image: "backgroundImage" },
+  });
+
+  const render = () => view.renderHoles(cellVm.holes, {
```
TheCodeSharman: logic esacped form the bindView again, violating the MVVM principles.

walta-app/app/views/SampleTrayEndcap.xml
```diff
@@ -0,0 +1,6 @@
+<Alloy>
```
TheCodeSharman: good to create new views for sub components thats fine.

TheCodeSharman:
As a general comment this isn't really meeting the north star goal, there are too many custom renders that do view realted titanium specific logic, these should all be replaced with mvvm/controllers and the goal is to get Titanioum sopecific logic to zero. (as we did with Menu and SampleTray)

walta-app/app/lib/mvvm/controllers/SampleTray.js
```diff
@@ -0,0 +1,45 @@
...
+    taxaSource: source,
+    topics: services.topics,
+    toDip: platform.convertSystemToDip,
+    toSystem: platform.convertDipToSystem,
+  });
+
+  const unbind = bindView(view, vm, {
+    tray: {
+      width: "trayWidthCss",
+      endcap: collection("endcapTiles", "SampleTrayEndcap"),
```
TheCodeSharman: isn't there only one end cap? do we need a colelction here??

walta-app/app/lib/viewmodels/SampleTray.js
```diff
@@ -0,0 +1,322 @@
...
+    this._onSourceChange = () => this.refresh();
+    if (typeof taxaSource.onChange === "function") taxaSource.onChange(this._onSourceChange);
+  }
+
+  // ── Geometry (bit-for-bit with the old controller's getters) ──────────────
+
+  // Titanium hands the raw view size (system px) via a measure() binding; convert
+  // to dip and re-derive geometry. Returns whether the reading was usable, so the
+  // binding retries until the view is actually laid out.
+  setViewport(size) {
```
TheCodeSharman, line 48: Is this why bindvierw needs emasure intrafstructure - is there a way to stop this leaking into the interface for bindView? It ok being int he implemntation of bindView but referenced as part of the bindView interface I'm not sure?

walta-app/app/lib/viewmodels/SampleTray.js
```diff
@@ -0,0 +1,322 @@
...
+      : null;
+  }
+}
+
+// A container of taxa cells (the endcap's 2, an interior tile's 4). Holds a stable
+// array of SampleTaxaIconViewModel slots — one per position, retained across kind
+// changes so a taxon icon is reused in place (the positional correctness the spec
+// pins). Structural + content changes flow through each slot's own notify; the
+// cell notifies only its own geometry (on a viewport change).
+class CellsViewModel extends ChangeNotifier {
```
TheCodeSharman: Shouldn't this `CellViewModel`, and `EndCapViewModel` belng in `viewmodel/SampleTrayTileViewModel` and/or `viewmodel/SampleyEncapViewModel`. Also I can't the `viewmodel/SampleTaxaIconViewModel` file. Oh! Just found it its in *this* file. We want them as sepreate viewmodel files, and consildate it to make a coherent domain model object oriented structure, currently you've got duplicates, I think it should be easy enough to simplify.

TheCodeSharman: Also don't forget unit tests for them.

walta-app/app/styles/SampleTaxaIcon.tss
```diff
@@ -24,7 +24,29 @@
...
+  height: "50%"
+}
+
 ".padIcon": {
   width: "80%",
   height: "80%"
 }
+
+// The add icon, sized + positioned to match the old plus button.
+".plusIcon": {
```
TheCodeSharman: I think the pkus icon is different enough tto be exttacted as a seperate mvvm componet: e.g. a mvvm/controller, viewmodel, and alloy view.

walta-app/app/lib/logic/SampleTraySource.js
```diff
@@ -0,0 +1,35 @@
...
+    name: keyTaxon ? keyTaxon.name : "unknown",
+  };
+}
+
+module.exports = function createSampleTraySource(taxaCollection, key, readonly) {
+  return {
+    length: function () { return taxaCollection.length; },
+    at: function (i) {
+      var m = taxaCollection.at(i);
+      return m ? toIconData(m, key) : undefined;
```
TheCodeSharman: Please remove the defensive programming - no need to swallow errors, assume the taxaCollection is the right shape please.

walta-app/app/lib/logic/SampleTraySource.js
```diff
@@ -0,0 +1,35 @@
+// Alloy-model reads for the ice-cube tray, behind the source-adapter seam the
+// SampleTrayViewModel is built from. Model access (allowed in the mvvm layer,
+// unlike Ti views) lives here rather than in the window shell, so the VM stays
+// Node-testable against a fake source. Unlike SampleHistorySource this is built
+// per-screen from the live `taxa` collection + the key (both per-sample), so the
+// Alloy shell constructs it and hands it to the screen controller.
+function toIconData(m, key) {
+  var taxonId = m.get("taxonId");
+  var keyTaxon = key && taxonId ? key.findTaxonById(taxonId) : null;
```
TheCodeSharman: again defensive programming alert: key will always be present, and taxonID will also always be present. Please remove the unnecessary conditional - if the caller doesn't supply the right things then the key.findTaxonById() will have an error, so the conditional checks obscure real code problems.

PR #383, opened 2026-07-31
WB-216: Full SampleHistory port — collection binding + first-class row

TheCodeSharman:
## Summary

The full SampleHistory port onto the View seam — and, through review, the pattern the rest of the app follows: a **framework-agnostic presentation DSL** where a screen controller declares bindings and never touches Titanium, plus the SampleHistory Alloy shell reduced to the `TopLevelWindow` floor.

The arc of the branch:

1. **Collection binding earns its place.** The old imperative row machinery (`buildAndBindRow`/`disposeRow`/`renderRows` + a hand-rolled diff, ~30 lines) became a declarative adapter.
2. **…then collapses to zero glue.** Prompted by review ("can a convention reduce the glue to 0, like Flutter?"), the per-screen adapter became a single declaration.
3. **The binder is injected, not imported.** `View` hands each controller a `bindView` pre-bound with its Titanium-side deps, so no controller wires a Ti dependency.
4. **The shell drops to the floor.** SampleHistory's data source moves to the services bag; its sync overlay becomes a self-owning modal; the shell is now `TopLevelWindow` + title + one anchor-bar button.

```js
// lib/mvvm/controllers/SampleHistory.js — the whole controller
module.exports = function ({ view, services, bindView }) {
  const { collection } = bindView;
  const vm = new SampleHistoryViewModel({ sampleSource: services.sampleSource, topics: services.topics });
  const unbind = bindView(view, vm, { sampleTable: { rows: collection("rows", "SampleHistoryRow") } });
  return { vm, dispose() { unbind(); vm.dispose(); } };
};
```

## The pieces

- **Zero-glue collection convention** — `collection(getter, "ComponentName")`. `bindView` synthesises the adapter (key = `item.key`, create via injected factory, dispose via the handle, TableView `setData` vs ScrollView add/remove by feature detection). The explicit-adapter form stays as the escape hatch for the scroll-windowed tray. The keyed-diff reconciler stays a pure, Node-testable function.
- **`makeBinder(createComponent, palette)`** — `View` builds one per controller (`controllerContext`) so the controller receives a ready binder. `Menu` drops its `palette` thread; `SampleHistory` drops its `createComponent` thread. Markers still import statically when a module-scope `BINDINGS` needs them (Academy's `call`).
- **First-class `SampleHistoryRow`** — binds its own row VM and fires `EDIT_SAMPLE` itself, so the list needs no per-row wiring. The per-row-click ownership (fixes dropped dispatch on reused/reordered Android rows) lives in the row.
[...]

PR #386, opened 2026-08-03
WB-218: Remove the bindView collection escape hatch

TheCodeSharman:
## Summary
- The explicit-adapter form of `collection()` — `collection(getter, adapter)` with `scrollEvent`/`onScroll`/`attach`/`detach`/`update` — was added in the SampleHistory port (#383) **in anticipation** of the scroll-windowed SampleTray.
- The SampleTray MVVM port (WB-191 / #385) deliberately uses the **convention** form instead: the shell keeps the Titanium scroll listener and feeds `vm.setScrollOffset()`, which notifies so the same keyed diff reconciles the visible window. So the escape hatch is now **unused across the entire app** — SampleHistory and SampleTray both use `collection(getter, "ComponentName")`.
- `collection()` now accepts only a component name; `setupCollection` drops the `attach`/`detach`/`scrollEvent`/`onScroll`/`update` handling and always synthesises the convention adapter; `validate()` drops the explicit-adapter check. **Kept**: the keyed-diff reconcile and the `render`/setData path — the convention uses both.
- The spec's reconcile-behaviour coverage (retain identity, unbind-disposes-all, render-mode reuse) **moves from explicit test adapters onto the convention form**, so the behaviour stays pinned rather than lost with the deleted tests.
- Doc: `docs/patterns/screen-controllers.md` updated — the convention is the only form; notes how a scroll-windowed list fits it.

A follow-up tidy surfaced by [Trello WB-218](https://trello.com/c/vkRf0g3W/218-remove-the-bindview-collection-escape-hatch). Independent of #385 — this branches off `main` (the escape hatch is already dead there) and touches only `bindView`; #385 doesn't modify `bindView`.

## Test plan
- [x] `npx grunt unit-test-node` — 463 passing (bindView collection binding + name-convention suites cover create / retain-identity / dispose / unbind / render-mode)
- [ ] SampleHistory device spec + acceptance stay green in CI (the convention's on-device consumer — behaviour-preserving, only dead branches removed)

### DIFF
--- a/docs/patterns/screen-controllers.md
+++ b/docs/patterns/screen-controllers.md
@@ -104,11 +104,6 @@ proxies) lives in the row, not the list.
 The row VM exposes a `key` getter — the identity the keyed diff reconciles on
 (the convention, mirroring a Flutter `ValueKey`).
 
-The explicit `collection(getter, adapter)` form — an object with
-`{ key, create, render, dispose }` — remains as the **escape hatch** for cases a
-name convention can't cover. Reach for it only there; anything the convention
-covers stays in the convention.
-
 ### Polymorphic lists and single fixed children
 
 Two arity variants cover the tray:
--- a/test/util/bindView_spec.js
+++ b/test/util/bindView_spec.js
@@ -505,66 +505,6 @@ describe("bindView collection binding", function () {
     set items(v) { this._items = v; this.notifyListeners(); }
   }
 
-  // The adapter is where the Titanium-specific work is injected: create() makes
-  // a child view, dispose() tears it down. bindView never touches Ti.
-  function makeAdapter() {
-    const disposed = [];
-    return {
-      disposed,
-      key: (item) => item.id,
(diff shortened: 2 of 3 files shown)


docs/patterns/screen-controllers.md
```diff
@@ -104,11 +104,11 @@ proxies) lives in the row, not the list.
 The row VM exposes a `key` getter — the identity the keyed diff reconciles on
 (the convention, mirroring a Flutter `ValueKey`).
 
-The explicit `collection(getter, adapter)` form — an object with
-`{ key, create, render, dispose }` — remains as the **escape hatch** for cases a
-name convention can't cover, i.e. things Titanium won't do natively: the
-scroll-windowed tray drives its window from `scrollEvent`/`onScroll`. Reach for it
-only there; anything the convention covers stays in the convention.
+`collection(getter, name)` is the only form — bindView synthesises the adapter
```
TheCodeSharman:
```suggestion
`collection(getter, name)` — bindView synthesises the adapter
```

PR #391, opened 2026-08-06
WB-190: inject the current sample, remove the global (PR-1)

TheCodeSharman:
## Summary
First slice of [Trello WB-190](https://trello.com/c/1TohSHqr) — **de-global the current sample** so the survey and (upcoming) training workflows can share the same components with different data + behaviour. This PR is the enabling substrate; it removes no behaviour.

- **Why:** every survey screen reaches `Alloy.Models.instance("sample")` / `Alloy.Collections.taxa`, and `SampleEditMenu` *reassigns* those globals. That single mutable singleton is why a training session can't coexist with a real survey, and why "identify a creature" is hard-wired to the survey path. Injecting the model is the prerequisite for a clean survey-vs-training split (`Survey` stays survey-only; training will supply its own model to the same slots — and omit EditTaxon).
- `SampleTraySource` no longer reads `Alloy.Models.sample` for `surveyType` — the sample is injected (and it's now Node-testable). `7a5e8e8d`
- `Navigation` gains `setCurrentSample()` and threads the sample+taxa into every screen's `args` in `onOpenView`, the same central mechanism already used for `key`/`Survey`. Additive — nothing reads `args.sample` yet. `7f279e61`

## Remaining on this branch (in progress)
- Model factory (`createNewSample`/`startNewSurveyIfComplete`) returns the sample instead of setting the global; the roots (new-survey, history view/edit) call `setCurrentSample`.
- Migrate consumers to read `args.sample`/`args.taxa`: SiteDetails, Habitat, Notes, LocationEntry, Summary, EditTaxon, the SampleTray shell, SampleEditMenu.
- Remove the global writes (SampleEditMenu reassign, index-app, AppReset).
- Keep EditTaxon a **survey-workflow** part (not baked into the shared tray) so training can wire a different on-identify step.

## Test plan
- [x] `npx grunt unit-test-node` — passes (incl. new SampleTraySource + Navigation-threading specs)
- [ ] Device unit specs (survey flow + identify flow) once consumers are migrated
- [ ] Manual: start survey → site details → tray → edit taxon, and history view/edit

### DIFF
--- a/test/logic/Navigation_sample_spec.js
+++ b/test/logic/Navigation_sample_spec.js
@@ -1,0 +1,34 @@
+require("mocha");
+const { expect } = require("chai");
+const { Navigation } = require("../../walta-app/app/lib/logic/Navigation");
+
+// The current sample/taxa thread through args the same way key/Survey already do
+// (see onOpenView), so screens receive the model by injection rather than reading
+// the Alloy.Models.sample global.
+describe("Navigation sample threading", function () {
+  let opened, services, nav;
+
+  beforeEach(function () {
+    opened = [];
+    services = {
+      Key: { url: "k" },
+      Survey: { name: "survey" },
+      View: { openView(ctl, args) { opened.push({ ctl, args }); return Promise.resolve(); } },
+    };
+    nav = new Navigation(services);
+  });
+
+  it("injects the current sample and taxa into an opened controller's args", function () {
+    const sample = { id: 1 };
+    const taxa = { length: 3 };
+    nav.setCurrentSample(sample, taxa);
+
+    nav.onOpenView("SiteDetails", { slide: "none" });
+
+    expect(opened[0].args.sample).to.equal(sample);
+    expect(opened[0].args.taxa).to.equal(taxa);
+    // still threads the existing app singletons
+    expect(opened[0].args.key).to.equal(services.Key);
+    expect(opened[0].args.Survey).to.equal(services.Survey);
+  });
+});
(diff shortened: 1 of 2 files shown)


walta-app/app/lib/mvvm/controllers/SampleTray.js
```diff
@@ -13,7 +13,7 @@ module.exports = function createSampleTray({ view, args, services, bindView }) {
   const { collection, component, input, measure, command, ref } = bindView;
   const platform = services.platform;
 
-  const source = SampleTraySource(Alloy.Collections["taxa"], args.key, args.readonly === true);
+  const source = SampleTraySource(Alloy.Collections["taxa"], args.key, args.readonly === true, Alloy.Models.instance("sample"));
```
TheCodeSharman, line 16: OK this is still a global ?

TheCodeSharman: Yes — interim scaffolding, and it needs to go. This commit only lifted the sample read up from SampleTraySource to the controller boundary; the taxa global was already here. End state is `SampleTraySource(args.taxa, args.key, args.readonly === true, args.sample)`, reading Navigation's threaded args. It's sequenced right after the model-factory + roots commit (a root has to call setCurrentSample before args.sample/taxa are populated, or the tray reads undefined). Removing it next — will link the commit here.

TheCodeSharman: Done in #393 (`5a2a21a6`): the tray controller now reads `args.taxa`/`args.sample` (threaded by Navigation off SURVEY_STARTED) instead of the globals. Device spec reworked to inject them through openArgs; iOS SampleTray 33/33.

PR #402, opened 2026-08-06
WB-190: TrainingRepository + waterbug_training store

TheCodeSharman:
## Summary
Persistence foundation for the training vertical (PR-3a) — and where the **`SampleTray` domain aggregate** is established greenfield.

Through review this grew from "a training store" into the domain spine the whole app revolves around:

- **`SampleTray`** (`lib/models/SampleTray.js`) — the ordered collection of `Taxon` the app is *about*. A pure, observable domain aggregate (extends `ChangeNotifier`), **decoupled from persistence and from presentation**. Training is a bare tray; a survey `Sample` will contain one. The survey migrates its Alloy `taxa` collection onto this later (that migration is the point — kill the persistence-coupled Backbone model that won't port to Flutter).
- **`Taxon`** (`lib/models/Taxon.js`) — a creature in the tray: `id` (stable verdict key), `taxonId`, `position`. Hand-written accessors, no schema coupling.
- **`TrainingRepository`** — a real repository, not a row-hash DAO: it **creates, persists, and returns `SampleTray`/`Taxon` domain models**. Dependency flows controller → repository → models; the models never call the repo (no fat model), and they never import `Ti.Database` (decoupled from the persist engine). It's the Flutter port seam — swap `Ti.Database` for `sqflite`/`drift` behind the same interface.

Other settled points:
- **Own isolated `waterbug_training` DB** — no photos/abundance/sync; can't collide with `loadCurrent`/sync/history by construction. (Training draft store; the richer survey/edit draft store is a separate follow-on.)
- **Persist + resume** — a session survives the OS reclaiming a backgrounded app (`currentSessionCode()` + `loadTray()` recover it). Cleared only on an explicit new exercise code or `clear()`.
- **Caller-supplied `position`** — `addTaxon(tray, taxonId, position)` stores the placement it's told; the append policy lives in the caller, leaving room for user-chosen cells later.
- Duplicate `taxonId`s stay distinct (a trainee's repeated misidentification must remain visible, not be deduped).

## API
`startSession(code) → SampleTray` · `currentSessionCode()` · `loadTray() → SampleTray` · `addTaxon(tray, taxonId, position) → Taxon` · `removeTaxon(tray, taxon)` · `clear()` · `close()`

## Test plan
- [x] `npx grunt unit-test-node` — 567 passing (6 new `models/SampleTray`, pure/Node-testable)
- [x] Device spec `TrainingRepository` — 10 passing (empty-tray start / add-returns-model / order-by-position / stable-id / duplicates / **resume-after-reopen** / new-session-clears / remove / clear)
- [ ] Full iOS device unit suite (pre-merge)
- [ ] Android emulator unit suite (CI)

Next: PR-3b — a training `SampleTraySource` adapting the `SampleTray` aggregate to the tray VM's source interface (`length`/`at`/`onChange`), so the existing `SampleTrayViewModel` presents it unchanged.

### DIFF
--- a/docs/patterns/repository-pattern.md
+++ b/docs/patterns/repository-pattern.md
@@ -173,30 +173,27 @@ via `Migrator.migrate(dbName)` at app startup. Repositories don't run
 their own migrations on open; that would couple every consumer of the
 repo to an implicit "first call sets up the schema" behaviour.
 
-## Which db: shared `waterbug_data`, or an isolated one
+## Shared db: `waterbug_data`
 
-Internal, low-stakes data (logs, a future cache) shares one SQLite db,
-`waterbug_data`. Lock contention isn't a concern at this app's scale
-(mobile, single user, low write volume), and a shared db keeps things
-simple.
+All non-Alloy tables — logs, training, a future cache — live in one
+SQLite db, `waterbug_data`. Lock contention isn't a concern at this
+app's scale (mobile, single user, low write volume), one
+`Migrator.migrate("waterbug_data")` at startup covers every table, and
+the migration manifest is global precisely because there's one db.
 
-**Isolated draft stores get their own db.** Training persists to
-`waterbug_training`, kept entirely apart from the real-sample archive
-so training data *can't* leak into sync/upload/history queries — those
-run against Alloy's `samples` db and structurally cannot see another
-file. Isolation-by-construction beats a discriminator column you have
-to remember to filter on. The survey/edit draft store will be a
-separate isolated db for the same reason.
-
-An isolated db is migrated with its own `Migrator.migrate(dbName)` call
-at startup (alongside the `waterbug_data` one). Note the current
-limitation: the migration manifest is global, so `migrate(dbName)`
-applies *every* migration to whichever db it's given — fine while the
-tables are harmless to co-create, but scoping the manifest per-db is
-the natural next step as more isolated dbs appear.
+This already isolates draft/training data from the real-sample
+**archive**: Alloy's `sample`/`taxa` use a *separate* db (`samples`),
+and every sync/upload/history query — `loadCurrent` and friends — runs
+against `samples`. A table in `waterbug_data` is structurally invisible
+to them, so training can't leak into sync/history without any
+discriminator-column filtering. Isolation comes from being a different
+db than `samples` — not from a per-store db file, which would only add
+a second `migrate` call and force per-db manifest scoping for no real
+gain.
 
 Alloy's `samples` and `taxa` databases stay separate per Alloy's
-convention.
+convention — Alloy hardcodes their `db_name`, so they'd only join
+`waterbug_data` if they ever migrate off Alloy.
 
 ## Adding a new repository
 
--- a/walta-app/app/lib/repository/TrainingRepository.js
+++ b/walta-app/app/lib/repository/TrainingRepository.js
@@ -1,9 +1,10 @@
 const SampleTray = require("../models/SampleTray");
 const Taxon = require("../models/Taxon");
 
-// Persists and hydrates the training SampleTray, in its own `waterbug_training`
-// DB — kept entirely apart from the real-sample archive so training data can
-// never leak into sync/upload/history. LogRepository-style: `open()` expects
+// Persists and hydrates the training SampleTray. Its tables live in the shared
+// non-Alloy `waterbug_data` DB — a *different* DB from Alloy's `samples`
+// archive, which is what every sync/upload/history query runs against, so
+// training data can never leak into them. LogRepository-style: `open()` expects
 // the schema to already exist (Migrator.migrate has run).
 //
 // This is the persistence seam: it creates the domain models (SampleTray/Taxon)
--- a/walta-app/app/spec/util/repository/TrainingRepository_spec.js
+++ b/walta-app/app/spec/util/repository/TrainingRepository_spec.js
@@ -5,7 +5,7 @@ var { removeDatabase } = require("spec/util/TestUtils");
 var TrainingRepository = require("repository/TrainingRepository");
 var Migrator = require("repository/Migrator");
 
-const TEST_DB = "waterbug_training_test";
+const TEST_DB = "waterbug_data_test";
 
 describe("TrainingRepository", function () {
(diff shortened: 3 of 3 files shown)


walta-app/app/lib/repository/migrations/202608070000000_training_session.js
```diff
@@ -0,0 +1,12 @@
+exports.up = function (migrator) {
+    migrator.createTable({
+        columns: {
+            "sessionCode": "TEXT PRIMARY KEY",
```
TheCodeSharman: do we need a seperate training_session table? note there wil only be one at a time, I think we can get away without this table

TheCodeSharman: Agreed — dropped it. And since we settled that training should **start fresh on relaunch** (not resume a killed session), the resume machinery it existed for is gone too:

- `training_session` table deleted (its `startedAt` column was never read anyway).
- Fresh-start is now encoded in the store itself: `open()` wipes stale rows, so a killed session never comes back.
- The session code lives in memory for the current run only, so the `sessionCode` column on `training_taxa` was also dead weight — dropped it. `training_taxa` is now just `(id, position, taxonId)`.

Net: one table, one migration, no dead columns. The resume-after-restart spec flipped to a fresh-start spec (start → add taxa → reopen → empty). All 8 device specs green.

walta-app/app/lib/repository/TrainingRepository.js
```diff
@@ -0,0 +1,77 @@
...
+            wipe();
+            sessionCode = code;
+        },
+
+        currentSession: function () {
+            return sessionCode;
+        },
+
+        addTaxon: function (taxonId) {
+            const position = nextPosition();
```
TheCodeSharman: How will this be referenced by the SampleTrayStore - doesn't it need to be indexed by position in the tray ?

walta-app/app/lib/repository/migrations/202608070000000_training_session.js
TheCodeSharman: Update after we talked it through: the table stays, but the whole store got reshaped. Landing points:

- **Persist + resume** is the requirement (recover an in-flight session after the OS reclaims a backgrounded app), so `training_session` is back as the resume target — a killed session is *not* meant to start fresh. Cleared only on an explicit new exercise code in Academy, or `clear()`.
- **`position` is now caller-supplied** — `addTaxon(taxonId, position)`. The store no longer computes `MAX(position)+1`; that append policy is the tray's business logic and lives in the caller (leaves room for user-chosen cells later). `id` stays the stable verdict key; `position` is the tray slot.
- Kept training in its **own isolated DB** (not the samples DB + a flag) so it can't collide with `loadCurrent`/sync/history by construction. It's a stripped-down store — no photos/abundance — which is the training/survey-draft split we agreed: separate simple store, small duplication in its favour.
- Repo stays a **bespoke Ti.Database layer** rather than an Alloy model, deliberately — it's the Flutter port seam and decouples domain from schema.

8 device specs green (incl. resume-after-reopen). Model + `TrainingTraySource` come in PR-3b.

TheCodeSharman: Final shape after the design discussion: this PR now also introduces the **`SampleTray` domain aggregate** — the ordered collection of `Taxon` the app revolves around, as a pure observable model decoupled from persistence, and the repo hydrates/persists it (returns domain models, not row hashes). The `training_session` table stays as the resume target. Layering is controller → repository → models; models never call the repo. 10 device specs + 6 Node specs green.

TheCodeSharman:
Full device unit suites (fresh build, no LiveView) — both green with the SampleTray aggregate + repo rework:

- **iOS: 413 passing, 0 failing**
- **Android: 408 passing, 0 failing** (15 pending = manual/skipped)

The `TrainingRepository` device spec (10) passes on both platforms; no cross-spec breakage (the `Topics.reset()` full-suite flake check is clean). Also updated `docs/patterns/repository-pattern.md` to distinguish a Repository (returns domain models) from a row-DAO (returns rows), since it previously presented LogRepository's row-hash shape as the template.

docs/patterns/repository-pattern.md
```diff
@@ -26,6 +37,56 @@ view updates and sync naturally. It's a poor fit for everything else:
...
+model the app mutates and observes?**
+
+| | Repository | row-DAO |
+|---|---|---|
+| Returns | domain model objects (aggregate/entity) | plain row objects `{ col: value }` |
+| For | data with behaviour + a live view (the tray) | internal append-only / query-only data (logs, a cache) |
+| Direction | controller → repository → models; models never touch the repo | caller → DAO → rows |
+| Example | `TrainingRepository` → `SampleTray`/`Taxon` | `LogRepository` → log rows |
+
+`LogRepository` returning row hashes is correct **because logs have no
```
TheCodeSharman, line 53: I think this extra detail in the documentation coudl be removed if we built a light weight domian log for the logs to use - this is cheap and it means there is one less way to confuse the repository pattern. This would be a follow up PR though, can you card it up.

walta-app/app/lib/models/Taxon.js
```diff
@@ -0,0 +1,16 @@
+// A creature placed in the SampleTray: a domain entity, decoupled from any
+// persistence engine. `id` is the stable identity (the tray's verdict key);
+// `position` is the tray slot; `taxonId` points at the taxon in the key.
+class Taxon {
```
TheCodeSharman, line 4: Does this need a change notifier too? Eventually when we migrate the survey flow to this it will I think, but YAGNI, lets put that off until its really needed.

walta-app/app/lib/repository/TrainingRepository.js
```diff
@@ -0,0 +1,81 @@
...
+                        id: rs.fieldByName("id"),
+                        taxonId: rs.fieldByName("taxonId"),
+                        position: rs.fieldByName("position")
+                    }));
+                    rs.next();
+                }
+            } finally {
+                rs.close();
+            }
+            return new SampleTray(taxa);
```
TheCodeSharman, line 56: Yes! Love it - proper domain model.

PR #406, opened 2026-08-07
WB-226: MethodSelect training mode — grey out non-key options

TheCodeSharman:
## Summary
- Training mode greys and disables Speedbug, Browse and Unknown bug in the MethodSelect modal, leaving only the Key selectable — matching the card mockup.
- Rewires MethodSelect onto a Titanium-free `MethodSelectViewModel` bound via `bindView`: a tap routes through the VM (which no-ops a disabled entry), and the VM's `*Disabled` getters drive the greying. Behaviour lives in the VM, appearance in the button.
- `MenuButton` grows a bindable `disabled` state (grey background + dim to 0.5 opacity); additive and default-off, so the other menus are unaffected.
- Selection→topic routing is Node-tested (`test/controllers/MethodSelect_spec.js` + `test/viewmodels/MethodSelect_spec.js`); the device spec drives the real modal and asserts the on-device greying.
- Drive-by (same presenter): dispose the `unknownbug` entry in `cleanUp()` — it was created but never cleaned when present.

Implements the screen for [Trello WB-226](https://trello.com/c/NK1YyxyU) — the MethodSelect training-mode greying. Threading `training: true` into the modal from the training tray's plus is part of the training-flow wiring (WB-190/PR-4), not this PR.

## Test plan
- [x] \`npx grunt unit-test-node\` — passes (579)
- [x] iOS device spec (\`--grep=MethodSelect\`) — 4 passing; \`--manual\` screenshot matches the mockup (Key highlighted, rest greyed)
- [ ] Android emulator device specs (CI)
- [ ] iOS/Android visual regression (CI, advisory)


walta-app/app/controllers/MethodSelect.js
```diff
@@ -1,35 +1,33 @@
...
 
+// Residual-Titanium presenter: builds the menu entries and exposes them. The
+// screen controller (lib/mvvm/controllers/MethodSelect) binds clicks + the
+// training-mode greying via bindView — no decision logic lives here.
 var size = "30.67%";
 if ( $.args.unknownBug ) {
    size = "25%"
 }
 
 $.keysearch = menuEntry( $.content, "/images/key-icon.png", "Key", null,
```
TheCodeSharman: We need to reduce hte legacy alloy controller to zero, and move the entire menu into the bindView MVVM framework. Could we make the menu entries a collection and resue the colelction binder for the menu??

TheCodeSharman: Agreed — moving the whole menu onto the collection binder. In progress on this branch:

- `MenuEntryViewModel` (new row-VM): `key`/`icon`/`title`/`description`/`disabled`/`select()`.
- `MethodSelectViewModel.entries` returns the ordered row-VMs.
- `MenuButton` becomes a bound component — a registered `lib/mvvm/controllers/MenuButton` binds the row-VM; the Alloy `MenuButton` reads `args.rowVm || args` so its one legacy caller (MayflyMusterSelect via `menuEntry`) is untouched.
- Screen controller: `content: { entries: collection('entries', 'MenuButton') }`; `controllers/MethodSelect.js` drops to an empty presenter shell.

TheCodeSharman: Done in bdebb63f. The menu is now a collection of `MenuButton` components bound via `content: { entries: collection('entries', 'MenuButton') }`; `controllers/MethodSelect.js` is an empty presenter shell. `MenuEntry` is the row-VM (icon/title/description/size/disabled/select), `MenuButton` reads `rowVm || args` so MayflyMusterSelect is untouched. Renders pixel-identical on-device (verified via `--manual` screenshot); greying flows row-VM → binder → button colour/opacity.

One note for your eye: the collection passes the row-VM as the component's only input, so presentation hints (card height, greyed colours) now live on `MenuEntry`. That matches how the other row-VMs carry display state; shout if you'd rather the greys move to Palette symbols.

walta-app/app/controllers/MenuButton.js
```diff
@@ -1,45 +1,61 @@
...
-if ( $.args.small ) {
-  $.resetClass( $.button, ["small"] );
-}
-if ( $.args.fill ) {
+// Layout: collection entries are full-width cards sized by the row-VM; legacy
+// callers pass small / fill / size as args.
+var fill = rowVm ? true : $.args.fill;
+var size = rowVm ? rowVm.size : $.args.size;
+
+if ( fill ) {
```
TheCodeSharman: Ok now we need to use `bindView` MVVM for MenuButton and reduce this file to zero as well.

TheCodeSharman: Done in 4fa8f18a. `controllers/MenuButton.js` is deleted outright — Alloy's generated controller plus the View seam's `destroy()`/`off()` fallback cover what the empty shell did, so there's nothing left to keep as a stub. The last imperative writes moved off it:

- **card height** → binds from the row-VM's `size` (`button: { height: "size", ... }`)
- **a11y label** → new `accessibilityLabel` getter (`title || description`); binding it is actually better than the old construction-time write, since bindView re-applies on first layout and dodges iOS's pre-open a11y-label drop
- **full-width** → baked into `MenuButton.tss` (`#button width: 96%`), and the now-unused `.large`/`.fill`/`.small` classes are gone, since every card is a full-width card

`lib/ui/MenuBuilder.js` (`menuEntry`) is deleted too — it had zero callers once the dead Mayfly UI was removed (#408). Verified: Node 606 green, and the MethodSelect device spec passes both contexts on a clean bundled build with the greying/full-width/varying-height layout confirmed on-device.

PR #411, opened 2026-08-07
WB-190 (4b): Add taxa, assess, and celebrate a clean run

TheCodeSharman:
## Summary
Second of three slices for the training vertical (Trello [WB-190](https://trello.com/c/1TohSHqr)). **Stacked on #410 (4a)** — base retargets to `main` when 4a merges; review the last commit for 4b's own diff. Wires the middle of the flow: identify taxa via the key, grade them, and open the success modal on a clean run.

- **Add-taxon** — `Training.addTaxon` appends an identified taxon to the session tray. `Main` routes `IDENTIFY` through the training tray when a session is active: a fresh identification (no `sampleTaxonId`) is added via the repository and the tray reopens **without a `taxonId`**, so the survey-only EditTaxon overlay never appears.
- **Assess** — the training tray shows an **Assess** anchor button (imperative like its GoBack/GoForward siblings, firing `Topics.ASSESS`; the tray VM listens and grades itself). Kept out of `bindView` deliberately — the anchor-bar → bindView migration is tracked in **WB-217**; doing it here would pull cross-screen nav refactoring into this PR.
- **Verdicts + Success** — on a clean run (every graded taxon correct) the VM emits `allCorrect`; the screen translates that to `Topics.TRAINING_SUCCESS` and `Main` opens the `TrainingSuccess` modal.
- **MethodSelect gating** — `SampleTrayPlus` flags `training` in its `SELECT_METHOD` intent so MethodSelect (from #406) greys the non-key options.

Rebased onto `main` after #406/#409 merged, so it composes with the merged MethodSelect gating.

Deferred to **4c**: the full capstone acceptance scenario, correcting a wrong taxon (tile remove/re-identify), and session teardown/resume on Finish.

## Test plan
- [x] `npx grunt unit-test-node` — 624 passing (Training.addTaxon, VM allCorrect + ASSESS subscription, controller → TRAINING_SUCCESS, SampleTrayPlus training flag)
- [x] iOS device specs `SampleTray`/`Academy`/`MethodSelect` — 41 passing on a clean build (new Assess-button test; survey path + MethodSelect gating unregressed)
- [ ] Cucumber capstone end-to-end — slice 4c

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #412, opened 2026-08-07
WB-190 (4c): Corrections, valid-code gating, capstone scaffolding

TheCodeSharman:
## Summary
Final slice for the training vertical (Trello [WB-190](https://trello.com/c/1TohSHqr)). **Stacked on #411 (4b)** — base retargets to `main` as the stack merges; review the later commits for 4c's own diff. Completes the training behaviours and lands the **green** capstone acceptance scenario.

Per the design decisions on this card:
- **Correct a taxon in place** — tapping a taxon in a training tray reopens MethodSelect and re-identifies it at the **same position** (`Training.beginReplace` + replace-aware `addTaxon`), so positional grading is preserved.
- **Valid-code gating** — the Academy **Start button is greyed/disabled until the code maps to a real exercise, turning green** when it does (`Training.isValidCode`, VM `startEnabled`/`startColor`). Grey reuses the existing disabled-menu-entry precedent — one palette `disabled = #e6e6e6`, `MenuEntry` now points at it too.
- **Test exercise 999 = [181, 179]** (shallow, shipped in prod), verdict a11y label, and the full capstone scenario + page objects (TrainingSuccess screen, training-add, Assess, verdict reader, re-identify).

## Real bugs the end-to-end run surfaced (and fixed)
- **Assessment never graded correct** — the key yields *string* taxonIds (`"181"`) but exercises are authored as *numbers* (`181`), and `TrainingAssessor` compared with strict `===`. Every taxon (incl. exercise 101) graded incorrect and a clean run was unreachable. Fixed by comparing as strings.
- **Capstone flake** — a decision/detail window slides in over ~200ms; Appium taps faster than a human, so a tap mid-animation hit a stale coordinate and was dropped. Fixed **deterministically** (no fixed delay, no retry): `base-screen.clickWhenStable` polls the element's `getLocation()` via `driver.waitUntil` until stationary, then taps. Documented the "poll for state, never delay, never retry-to-go-green" rule in the tdd + fast-iteration skills.

Capstone runs **green 3/3** locally: Academy → 999 → empty tray → add gastropod → add wrong limpet → Assess → cross → re-identify to mussel → Assess → TrainingSuccess → Finish → menu.

Deferred (noted): the literal "offline" Given (training makes zero network calls); Finish teardown + resume (coupled — `startSession` wipes on each start, so no stale-data bug without resume).

## Test plan
- [x] `npx grunt unit-test-node` — 630 passing (Training corrections/validity/assessor, Academy VM, MenuEntry)
- [x] iOS acceptance "A valid session code enables Start" — passes
- [x] iOS acceptance capstone "Complete a training exercise, correcting a mistake" — passes 3/3

### DIFF
--- a/walta-app/app/styles/TrainingSuccess.tss
+++ b/walta-app/app/styles/TrainingSuccess.tss
@@ -27,15 +27,27 @@
 	font: { fontFamily: "Tahoma", fontWeight: "bold", fontSize: "38dp" }
 }
 
+// Sized by height (width SIZE keeps the aspect) so the whole modal — titlebar,
+// message, graphic, button — fits the shortest landscape viewport without
+// scrolling. Scales up with the screen-height buckets; the base serves the
+// smallest screens.
 ".successGraphic": {
-	top: "20dp",
-	width: "38%",
-	height: Ti.UI.SIZE
+	top: "12dp",
+	width: Ti.UI.SIZE,
+	height: "44dp"
+}
+
+".successGraphic[if=Alloy.Globals.isHighRes]": {
+	height: "64dp"
+}
+
+".successGraphic[if=Alloy.Globals.isXHighRes]": {
+	height: "120dp"
 }
 
 ".finishButton": {
-	top: "24dp",
-	bottom: "24dp",
+	top: "12dp",
+	bottom: "12dp",
 	color: Alloy.CFG.colors.primary,
 	backgroundColor: Alloy.CFG.colors.white,
 	borderColor: Alloy.CFG.colors.primary,


features/academy.feature
```diff
@@ -2,10 +2,25 @@ Feature: Academy
 
 I want to test my waterbug identification skills by starting a training session
 
-Scenario: Enter a training session code
+Scenario: A valid session code enables Start
```
TheCodeSharman: I don't think this scenario is necesary anymore, its covered and more by "Complete a training exercise, correcting a mistake"

walta-app/app/lib/logic/Training.js
```diff
@@ -19,13 +20,34 @@ module.exports = function createTraining({ topics, repo, exercises }) {
...
     isActive() {
       return tray !== null;
     },
 
-    // Append an identified taxon to the session tray. Position is the caller's
-    // concern (the store takes it as given), so training appends at the end.
+    // Mark a taxon for re-identification: the next addTaxon replaces it in place
+    // (the trainee tapped it to correct a wrong pick). See addTaxon.
+    beginReplace(sampleTaxonId) {
+      pendingReplace = sampleTaxonId;
```
TheCodeSharman: I think this is a code smell, it relies on recording some state, but that could potentially cause bugs if the user jumps around in the app a lot adn this state isn't reset, I'm wodnering if it make for sense to pass the tray position into the key identification and when the a taxon is identified, return that value. This also aligns with our decision to based the sample tray colecton on position.

features/academy.feature
TheCodeSharman: Done — removed the "A valid session code enables Start" scenario (and its now-orphaned step defs). The capstone covers entering 999 + Start becoming available, and the disabled-until-valid gating is covered by the Academy VM unit specs. Commit 54e60e1a.

walta-app/app/lib/logic/Training.js
TheCodeSharman: Reworked exactly as you suggested. The tapped cell's tray position now threads through MethodSelect → the key → the completing IDENTIFY, and `addTaxon(taxonId, position)` replaces that slot in place — no `beginReplace`/`pendingReplace` session state, so navigating away mid-correction can't leave a stale target. Aligns with the position-based tray collection. Commit 205ec2d7 (Node specs pin the threading end-to-end; iOS capstone green 16/16).

PR #427, opened 2026-08-14
WB-222: Give logs a LogEntry model; collapse to one repository shape

TheCodeSharman:
## Summary
- **One repository shape.** `LogRepository.query()` used to return plain row hashes (a row-DAO) while `TrainingRepository` returns domain models. That split forced a Repository-vs-row-DAO decision on every table and a page of docs to explain it. Now a repository always returns domain models.
- New `lib/models/LogEntry.js` — a deliberately **anemic** domain model (getters over `ts/level/facility/message`, no behaviour). The anemic class is the accepted cost of removing the design-space split: no DB schema leaking into callers, and somewhere to hang behaviour/change-notification if logs ever grow.
- `LogRepository.query()` returns `LogEntry` instances.
- `SyncFeedback` VM wraps the live `Logger` entries it appends, so the Show Logs pane holds one shape (`LogEntry`) whether an entry came from the repository or arrived live. Consumers already read `.ts/.level/.facility/.message`, which the getters satisfy unchanged (`DiagnosticsBundle`, the log pane) — no consumer edits beyond the wrap.
- **Docs: removed the "two shapes: Repository vs row-DAO" section, decision table, and per-table "which shape?" guidance** from `repository-pattern.md` (+ the README index line). This is the design-space/documentation reduction the card was really about.

Completes [Trello WB-222](https://trello.com/c/uWJzuthf). Per your call: implement as carded, one Repository pattern, anemic model accepted.

## Test plan
- [x] `npx grunt unit-test-node` — 633 passing (adds `LogEntry` model spec)
- [x] iOS device specs `LogRepository | SyncFeedback | DiagnosticsBundle` — 22 passing (runtime `instanceOf LogEntry` + consumers confirmed)
- [ ] CI green both platforms

### DIFF
--- a/walta-app/app/lib/models/LogEntry.js
+++ b/walta-app/app/lib/models/LogEntry.js
@@ -1,7 +1,4 @@
-// A single persisted log line: a domain entity, decoupled from any
-// persistence engine, mirroring the other repository-backed models.
-// Deliberately anemic — logs carry no behaviour — so the whole app uses
-// one repository shape rather than a Repository/row-DAO split.
+// A single persisted log line: a domain entity
 class LogEntry {
   constructor({ ts, level, facility, message }) {
     this._ts = ts;


walta-app/app/lib/models/LogEntry.js
```diff
@@ -0,0 +1,19 @@
+// A single persisted log line: a domain entity, decoupled from any
+// persistence engine, mirroring the other repository-backed models.
+// Deliberately anemic — logs carry no behaviour — so the whole app uses
+// one repository shape rather than a Repository/row-DAO split.
```
TheCodeSharman:
```suggestion
// A single persisted log line: a domain entity
```

PR #429, opened 2026-08-17
WB-242: Adding a Taxon goes back to empty tray

TheCodeSharman:
## Summary
- **Root cause:** `Training.isActive()` read `tray !== null`, which was never reset. After one Academy session, Main's IDENTIFY router treated every *later survey* identification as training — adding to the stale training tray and opening an empty `SampleTray` — so a survey's "Add to sample" left the tray empty.
- **Fix:** thread a `training` flag through the identification flow (MethodSelect payload → key/browse/speedbug navigation → `TaxonDetails` IDENTIFY, plus the tray re-identify tap) and route on `data.training`. Training-ness now travels with the identification instead of sitting in mutable session state a finished session can leak into a survey. Removes the now-dead `isActive()`.
- **Coverage:** new cucumber scenario in `sample_a_site.feature` — a full training session followed by a survey that still records identifications. Before the fix it failed with `EditTaxonScreen not present`.

Fixes [Trello WB-242](https://trello.com/c/UeqOKh9l/242-adding-an-taxon-goes-back-to-empty-tray) — first of the four training-state-leak defects; WB-244/245 (the deeper anchor-bar training-mode gating) build on this threading and land separately.

## Test plan
- [x] `npx grunt unit-test-node` — 631 passing (payload contract updated for the threaded flag)
- [x] `acceptance-test` (iOS sim) — new "survey after a training session" scenario RED before / GREEN after; existing "Sample collection" still green

### DIFF
--- a/features/sample_a_site.feature
+++ b/features/sample_a_site.feature
@@ -17,27 +17,6 @@ Scenario: Sample collection
    And a sample id is automatically created for the user
    And a sample is stored and sample tray is cleared
 
-# A training session leaves the training service holding a session tray. That
-# state must not leak into a subsequent real survey: before the fix, the survey's
-# "Add to sample" was routed as a training identification, so the taxon editor
-# never opened and the tray stayed empty.
-Scenario: A survey after a training session still records identifications
-  Given I am logged in as "test@example.com"
-    And the GPS has a fix
-    And a user has arrived at a site to sample
-  When I open the Academy from the menu
-   And I start the training session "999"
-   And I identify a gastropod through the key
-   And I identify a freshwater limpet through the key
-   And I assess the training tray
-   And I re-identify the limpet as a mussel
-   And I assess the training tray
-   And I finish the training
-   And the user fills out the site details
-   And the user fills out the habitat screen
-   And the user identifies a number of taxa
-  Then the sample tray is filled with each identification
-
 @skip
 Scenario: Cancel sample
   Given the user has identified a number of taxa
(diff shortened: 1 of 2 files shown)


features/sample_a_site.feature
```diff
@@ -17,6 +17,27 @@ Scenario: Sample collection
...
+# state must not leak into a subsequent real survey: before the fix, the survey's
+# "Add to sample" was routed as a training identification, so the taxon editor
+# never opened and the tray stayed empty.
+Scenario: A survey after a training session still records identifications
+  Given I am logged in as "test@example.com"
+    And the GPS has a fix
+    And a user has arrived at a site to sample
+  When I open the Academy from the menu
+   And I start the training session "999"
+   And I identify a gastropod through the key
```
TheCodeSharman: This doesn't seem like an acceptance test - I think we should write this as an end to end test instead becasue its a regfression test against an bug rather than an acceptance scenario. Thats why we have end to end tests: acceptance tests are for requirements from the users point of view, they define the minimal set of steps for a product to be accepted.

TheCodeSharman: Agreed — moved it to the end-to-end layer in 05e1bb76. Removed the acceptance scenario and added `end-to-end-testing/survey-after-training-test.js`, which drives the same flow (complete a training session, then run a full survey and assert the tray records each identification). Verified green on the iOS simulator via `end-to-end-test`.

PR #430, opened 2026-08-17
WB-244: Edit of a preexisting sample shows training-mode Method Select

TheCodeSharman:
Stacked on #429 (WB-242). Review the WB-244 commits; base will retarget to `main` once #429 merges.

## Summary
- **Root cause:** `Navigation` held `currentTraining`/`currentTray`/`currentAssessor` as ambient state, set on `TRAINING_STARTED` and only cleared on `SURVEY_STARTED`. Nothing cleared it when a training session *ended*, so `onOpenView` re-injected the stale flag into any survey screen reached **without** a fresh `SURVEY_STARTED` — e.g. an anchor-bar jump to Home and back into a survey (the app lets you jump around freely). The survey tray's **+** then opened the greyed, training-mode MethodSelect.
- **Fix (per the design discussion):** remove the ambient training state entirely. Training mode is now a **threaded parameter** — like a URL query on a stateless (HTTP-style) navigation. The Academy button and the training identify flow thread `training: true`; transitions that leave training (Home, a survey) simply don't carry it, so there's no variable to go stale and jumping around can't resurrect it. The training session's tray/assessor live in their owner, the `Training` service (`currentTray`/`currentAssessor`), threaded into the tray by the screens that open it.
- **Coverage:** fast Node `Navigation` tests pin the no-leak behaviour (a screen opened without a training flag is never training, even after a training screen / `TRAINING_STARTED`). New survey **key-path** identify scenario guards the methods stay enabled in a survey (browse is already covered by "Sample collection"). Academy + survey-after-training acceptance stay green.

**Deferred:** a speedbug-path survey scenario — `speed_bug.feature` is entirely `@skip` and needs a reliable driver + known refIds before it can be revived; that's a separate coverage effort, not a WB-244 blocker.

Follow-up carded as **WB-246**: thread the remaining ambient `currentSample`/`currentTaxa` too, converging on a single threaded navigation-context object.

## Test plan
- [x] `npx grunt unit-test-node` — 632 passing (incl. stateless `Navigation` tests)
- [x] `acceptance-test` (iOS sim) — "Complete a training exercise" + "survey after a training session" + "Identify a creature via the key in a survey" green

### DIFF
--- a/walta-app/app/lib/logic/Navigation.js
+++ b/walta-app/app/lib/logic/Navigation.js
@@ -9,15 +9,9 @@ function Navigation(services) {
     this.services = services;
     this.currentSample = null;
     this.currentTaxa = null;
-    // The roots that establish a survey's sample announce it on the bus; seed the
-    // sample threaded into every screen's args from there (see onOpenView).
+    // Seed currentSample from the survey root that announces it on the bus.
     services.topics.subscribe(services.topics.SURVEY_STARTED,
         (data) => this.setCurrentSample(data.sample, data.taxa));
-    // Training mode is NOT held here: it rides through each transition as an args
-    // parameter (like a URL query), so a screen is training-mode only when opened
-    // with training:true. A finished session can't leak into a later survey — there
-    // is no state to go stale — and the training session's tray/assessor live in
-    // their owner, the Training service, looked up by the screens that open it.
 }
 
 Navigation.prototype.getHistory = function () {
@@ -32,9 +26,7 @@ Navigation.prototype.setCurrentSample = function(sample, taxa) {
     this.currentTaxa = taxa;
 }
 
-// The survey sample/taxa are threaded in from the single active survey; training
-// mode and the training session's tray/assessor are NOT injected here — they ride
-// through in the caller's args so each screen keeps the context it was opened in.
+// Thread the active survey's sample/taxa into every screen's args.
 Navigation.prototype.onOpenView = function(ctl,args) {
     Object.assign(args, {
         key: this.services.Key,
--- a/walta-app/app/lib/logic/Training.js
+++ b/walta-app/app/lib/logic/Training.js
@@ -1,11 +1,7 @@
 const createTrainingAssessor = require("logic/TrainingAssessor");
 
 // Training-session service — the training-mode counterpart to Survey. Owns the
-// active session's SampleTray + assessor (per-session state, so a factory with a
-// closure rather than Survey's stateless module object). The screens that open the
-// training tray look the session up here (currentTray/currentAssessor) and thread
-// it on in their args; training mode itself is a threaded parameter, not state held
-// by Navigation. Titanium-free.
+// active session's SampleTray + assessor. Titanium-free.
 module.exports = function createTraining({ repo, exercises }) {
   let tray = null;
   let assessor = null;
@@ -29,11 +25,9 @@ module.exports = function createTraining({ repo, exercises }) {
       return exercises.loadExercise(code) !== null;
     },
 
-    // Add an identified taxon. With no position it appends (position = current
-    // length — the store takes position as given). Given a position, it drops the
-    // taxon already in that slot and puts the new one there, so re-identifying a
-    // wrong pick preserves positional grading. The position rides in from the tap
-    // through the key identification rather than being held as session state.
+    // Add an identified taxon. With no position it appends. Given a position, it
+    // replaces the taxon already in that slot, so re-identifying a wrong pick
+    // preserves positional grading.
     addTaxon(taxonId, position) {
       if (position == null) {
         return repo.addTaxon(tray, taxonId, tray.length);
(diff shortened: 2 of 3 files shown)


walta-app/app/lib/logic/Navigation.js
```diff
@@ -9,16 +9,15 @@ function Navigation(services) {
...
-    this.currentTray = null;
-    this.currentAssessor = null;
     // The roots that establish a survey's sample announce it on the bus; seed the
     // sample threaded into every screen's args from there (see onOpenView).
     services.topics.subscribe(services.topics.SURVEY_STARTED,
         (data) => this.setCurrentSample(data.sample, data.taxa));
-    // Training's counterpart: a training session announces its tray/assessor here.
-    services.topics.subscribe(services.topics.TRAINING_STARTED,
-        (data) => this.setCurrentTraining(data.tray, data.assessor));
+    // Training mode is NOT held here: it rides through each transition as an args
```
TheCodeSharman: PLease remove this comment - it talks about the previous code whioch makes no sense becasue the reader doesn't hae a mental model of all our mistakes.

walta-app/app/lib/logic/Training.js
```diff
@@ -2,10 +2,11 @@ const createTrainingAssessor = require("logic/TrainingAssessor");
 
 // Training-session service — the training-mode counterpart to Survey. Owns the
 // active session's SampleTray + assessor (per-session state, so a factory with a
-// closure rather than Survey's stateless module object), and announces it on
-// TRAINING_STARTED the way Survey announces SURVEY_STARTED. Navigation seeds the
-// tray/assessor it threads into every screen's args from that event. Titanium-free.
-module.exports = function createTraining({ topics, repo, exercises }) {
+// closure rather than Survey's stateless module object). The screens that open the
```
TheCodeSharman: Again same problem a comment defending a design decision - it makes no sense in the present with the code way way it is.

walta-app/app/lib/logic/Training.js
```diff
@@ -15,10 +16,14 @@ module.exports = function createTraining({ topics, repo, exercises }) {
       if (!order) return false;
       tray = repo.startSession(code);
       assessor = createTrainingAssessor(order);
-      topics.fireTopicEvent(topics.TRAINING_STARTED, { tray, assessor, training: true });
       return true;
     },
 
+    // The active session's tray + assessor — the training screens thread these into
```
TheCodeSharman, line 18: This is the right level of detail. CAn you update the coding sytle documetnation to make it clear NOT to write the bad type of comments. Comments should be short. They shoudl state what the code does. They shouldn't refer to wha thte code might do or did do.

walta-app/app/lib/mvvm/controllers/Academy.js
```diff
@@ -20,15 +20,19 @@ module.exports = function createAcademyController({ view, close, services, bindV
   const vm = new AcademyViewModel({ isValidCode: (code) => services.Training.isValidCode(code) });
   const unbind = bindView(view, vm, BINDINGS);
 
-  // Starting a session launches training for the entered code (loads the
-  // exercise, opens a session, announces the tray on the bus). On a known code we
-  // dismiss the modal and open the training tray — Navigation threads the session's
-  // tray/assessor from the TRAINING_STARTED it just fired. An unknown code is a
-  // no-op, leaving the modal up.
+  // Starting a session launches training for the entered code (loads the exercise,
+  // opens a session). On a known code we dismiss the modal and open the training
```
TheCodeSharman: Same problem again too wordy - cosnider if the comment is really adding anything - its obiouvs training is true becasue the code says it

walta-app/app/lib/logic/Navigation.js
TheCodeSharman: Removed — the constructor now just seeds currentSample from the bus, no commentary on what training mode is not. (23376382)

walta-app/app/lib/logic/Training.js
TheCodeSharman: Trimmed the header to what the module is (training-session service, owns the tray+assessor, Titanium-free); dropped the closure-vs-module and not-Navigation-state asides. (23376382)

walta-app/app/lib/mvvm/controllers/Academy.js
TheCodeSharman: Cut to two lines stating what happens on a known vs unknown code; dropped the training:true narration and the not-Navigation-state aside. (23376382)

walta-app/app/lib/logic/Training.js
TheCodeSharman: Agreed. Left this one as-is since it's the good kind. I'll codify the rule in docs/coding-style.md (short; state what the code does; don't describe what it did or might do) — doing that as its own small doc PR rather than bundling a methodology change into this fix. Will link it here.

TheCodeSharman: Doc rule is up as #434.

PR #431, opened 2026-08-17
WB-245: Bypass of training restrictions via the anchor bar

TheCodeSharman:
Stacked on #430 (WB-244) → #429 (WB-242). Review the WB-245 commit; base retargets down the stack as each merges.

## Summary
- **Defect:** in a training assessment only the key is allowed — MethodSelect greys speedbug/browse/unknown-bug — but the **speedbug/browse shortcuts on the KeySearch and TaxonDetails anchor bars** slip straight past that, so a user can jump into speedbug from a key question and cheat the exercise.
- **Fix:** omit those two shortcuts when the key/taxon screens are opened in **training mode**, using the `training` flag already threaded through the flow (WB-242/244) — no global state. Both screens are reachable during a training identify, so both are gated. Also gave the shortcuts **accessibility labels** ("Speedbug"/"Browse") — they were unlabelled icon buttons, an a11y gap, and it makes them locatable in tests.
- **Coverage:**
  - Device specs (`KeySearch_spec`, `TaxonDetails_spec`): the anchor bar **offers** the shortcuts in a survey and **hides** them in training (asserts on the anchor-bar tool images).
  - Acceptance (`academy.feature`): "Training keeps the key as the only identification path" — starts a session, begins a key identification, and checks the speedbug/browse shortcuts are gone.

The gating is a UI-composition concern, so the device specs are the primary coverage; the acceptance scenario is the end-to-end sanity check.

## Test plan
- [x] Device unit specs (iOS sim) — KeySearch + TaxonDetails gating green (15 passing)
- [x] `acceptance-test` (iOS sim) — "Training keeps the key as the only identification path" green

### DIFF
--- a/features/academy.feature
+++ b/features/academy.feature
@@ -16,13 +16,3 @@ Scenario: Complete a training exercise, correcting a mistake
   Then the training success screen is shown
   When I finish the training
   Then the menu is shown
-
-# The key is the only identification path allowed in training — the anchor bar must
-# not offer the speedbug/browse shortcuts, which would otherwise slip past the
-# greyed Method Select and defeat the exercise.
-Scenario: Training keeps the key as the only identification path
-  When I open the Academy from the menu
-  And I start the training session "999"
-  Then an empty training tray is shown
-  When I begin identifying a creature via the key
-  Then the speedbug and browse shortcuts are not offered
--- a/end-to-end-testing/training-key-only-path-test.js
+++ b/end-to-end-testing/training-key-only-path-test.js
@@ -1,0 +1,27 @@
+'use strict';
+// Regression: the key is the only identification path allowed in training. The
+// anchor bar must not offer the speedbug/browse shortcuts, which would otherwise
+// slip past the greyed Method Select and defeat the exercise. This drives the
+// real screen stack to prove the shortcuts are absent, so it lives in the
+// end-to-end layer rather than the business-readable acceptance suite.
+
+async function beginTrainingIdentifyViaKey(world) {
+    await world.menu.selectAcademy();
+    await world.academy.enterCode("999");
+    await world.academy.waitForStartAvailable();
+    await world.academy.start();
+    await world.sample.waitFor();   // the empty training tray
+    await world.sample.selectAddSample();
+    await world.methodSelect.viaKey();
+}
+
+describe('E2E: training keeps the key as the only identification path', function () {
+    this.timeout(180000);
+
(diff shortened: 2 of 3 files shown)


features/academy.feature
```diff
@@ -16,3 +16,13 @@ Scenario: Complete a training exercise, correcting a mistake
   Then the training success screen is shown
   When I finish the training
   Then the menu is shown
+
+# The key is the only identification path allowed in training — the anchor bar must
+# not offer the speedbug/browse shortcuts, which would otherwise slip past the
+# greyed Method Select and defeat the exercise.
+Scenario: Training keeps the key as the only identification path
```
TheCodeSharman: Same as the other PRs - I think this is an end to end test not an acceptance test. Accpetance tests are scneraios played out byut e user in acceptance workflow. This is a specific regression test so its doesn't belong as a regression test. e.g. the distinction is: would the client offer this as a buisness requiement? I'd argue no. This is a specific implementation of the current design not an acceptance criteria

walta-app/app/controllers/KeySearch.js
```diff
@@ -34,8 +34,13 @@ $.TopLevelWindow.addEventListener('close', function cleanUp() {
 var acb = $.getAnchorBar(); 
 $.args.name = "decision";
 var goBackBtn = Alloy.createController("GoBackButton", {slide: "left"} );
-acb.addTool( acb.createToolBarButton( '/images/icon-speedbug-white.png', Topics.SPEEDBUG, null, { surveyType: $.args.surveyType, allowAddToSample: $.args.allowAddToSample, position: $.args.position, training: $.args.training } ), true );
-acb.addTool( acb.createToolBarButton( '/images/icon-browse-white.png', Topics.BROWSE, null, { surveyType: $.args.surveyType, allowAddToSample: $.args.allowAddToSample, position: $.args.position, training: $.args.training }  ), true );
+// During a training assessment the key is the only allowed path, so omit the
+// speedbug/browse shortcuts — otherwise the anchor bar slips past the greyed
```
TheCodeSharman, line 38: I like this comment - its short. Its present tense. And it explains why.

features/academy.feature
TheCodeSharman: Moved to the e2e layer in 334354c9 — removed the scenario + its two step defs, added `end-to-end-testing/training-key-only-path-test.js` (the `shortcutPresent` screen helper stays shared). Green on the iOS simulator. Same treatment as #429.

PR #432, opened 2026-08-17
WB-243: Blank summary when editing a sample

TheCodeSharman:
Independent of the training-cluster PRs (#429/#430/#431); branched off `main`.

## Summary
- **Defect:** opening a completed sample in **Edit** mode and moving to the Summary showed a blank screen — no date, SIGNAL score or assessment. (Fine in View mode.)
- **Root cause:** the Summary view bound its fields to the `sample` **singleton** via `<Model src="sample"/>` (`{sample.score}`, `{sample.dateCompleted}`, …). In edit mode the sample being reviewed is a **temp copy** threaded through the flow as `args.sample`, not the singleton — so the summary rendered the singleton's stale/blank data. This surfaced when the draft-tray work moved screens onto args-threading; the Summary was never migrated.
- **Fix:** render the summary fields from `$.args.sample` in the controller. `transform()` already resolves the edit-mode date from `overrideDateCompleted` and the score from the copy's own taxa, so the edited sample renders correctly. Gave the previously data-bound labels ids and dropped the singleton `<Model>` binding.
- **Coverage:**
  - Device spec (`Summary_spec`): renders the SIGNAL score from the sample the controller is **given**, not the singleton (the old spec always passed `args.sample === singleton`, which is why it never caught this).
  - Acceptance (`view_samples.feature`): completes a survey, reopens it from history in Edit, walks forward to the Summary and checks it is populated (site info + SIGNAL Score).

## Test plan
- [x] Device unit spec (iOS sim) — `Summary controller` 26 passing incl. the new edit case
- [x] `acceptance-test` (iOS sim) — "Editing a completed sample shows a populated summary" green

### DIFF
--- a/features/view_samples.feature
+++ b/features/view_samples.feature
@@ -2,22 +2,6 @@ Feature: View samples
 
 I want to review the history of the samples I have already collected.
 
-# Regression guard for WB-243: the Summary was blank when reviewing a sample in
-# Edit mode, because the view bound to the global singleton rather than the edited
-# copy threaded through the flow.
-Scenario: Editing a completed sample shows a populated summary
-  Given I am logged in as "test@example.com"
-    And the GPS has a fix
-    And a user has arrived at a site to sample
-  When the user fills out the site details
-   And the user fills out the habitat screen
-   And the user identifies a number of taxa
-   And the user marks the sample as complete
-   And a sample is stored and sample tray is cleared
-   And I edit the stored sample from history
-   And I move forward to the summary screen
-  Then the summary shows the survey site and signal score
-
 @skip
 Scenario: Review history without server login
   Given I do not have a server login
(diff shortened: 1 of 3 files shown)


features/view_samples.feature
```diff
@@ -2,6 +2,22 @@ Feature: View samples
 
 I want to review the history of the samples I have already collected.
 
+# Regression guard for WB-243: the Summary was blank when reviewing a sample in
+# Edit mode, because the view bound to the global singleton rather than the edited
+# copy threaded through the flow.
+Scenario: Editing a completed sample shows a populated summary
```
TheCodeSharman: Again same problem this isn't an acceptance test - its is testing a specifc regression please make it an end to end test see other PR review comments for rationale.

TheCodeSharman: Moved to the e2e layer in 241bb250 — removed the scenario + its three step defs, added `end-to-end-testing/edit-sample-summary-test.js`, which drives the full create → store → edit → summary path and asserts the summary shows the site and SIGNAL score. Green on the iOS simulator. (The relocated version also drops the WB-243 reference from the comment, per the rule in #434.)

PR #435, opened 2026-08-19
WB-248: walta://reset leaks the in-progress survey into the next scenario

TheCodeSharman:
## What

`walta://reset` (the per-scenario iOS soft reset) now **re-seeds Navigation's in-memory current sample**, so a survey left mid-flight by one scenario can't leak into the next.

## Root cause

Diagnosed from a failed CI run's artifact (screenshot + page-source + device log):

1. On a contended runner the WebDriverAgent session drops mid-scenario (the app stays healthy — only the automation session blips). This leaves the app **mid-survey with a dirty sample**.
2. `walta://reset` wipes the sample DB **but** `Navigation.currentSample` is an *in-memory* reference, seeded once via `SURVEY_STARTED` and never re-read from the DB. Reset never clears it, so it keeps the previous scenario's dirty sample.
3. Reset's `Topics.HOME` → `garbageCollectControllers` unloads the dirty `SiteDetails` → `onDiscardEdits` → native **"Unsaved Changes"** modal.
4. `softResetInApp` only polls for the logged-out "Log In" button, so it hangs behind that modal (*"did not return to the logged-out menu"*).
5. The modal **survives the reset**, wedging every following scenario. Those cascade failures leave the session *alive*, so they aren't marked `session-dead`; `CucumberLauncher`'s retry only fires when *every* failure is infra (`infra.length === failedCount`) → `1 ≠ 6` → no retry → job red.

This flake hits `main` directly (blocked #430 from auto-merging).

## The fix

Re-seed a fresh, clean sample in `AppReset.reset()` by firing `SURVEY_STARTED` with the re-instantiated models — **exactly as `index-app` does at startup** — so the in-memory pointer matches the wiped DB and reset navigates home cleanly. No dialog is ever raised. `SURVEY_STARTED` is subscribed only by Navigation, so the re-fire is a targeted re-seed with no other side effects.

Not a production concern: an in-progress survey *is* persisted to the DB incrementally (`SiteDetails` saves each step; taxa save on add) and resumed at startup — so this only re-syncs the test reset's in-memory state with the DB it just wiped.

## Verification

Deterministic local repro (temp 2-scenario feature, iOS sim): scenario 1 leaves a survey dirty, scenario 2's `Before`-reset must recover.
- **Without fix:** scenario 2 fails — *"app did not return to the logged-out menu after reset"* (1 passed, 1 failed).
- **With fix:** 2 scenarios (2 passed), 17/17 steps — no dialog raised.

Plus a device unit spec (`AppReset_spec.js`) locking that `reset()` re-fires `SURVEY_STARTED` with a clean sample.

### DIFF
--- a/walta-app/app/lib/util/AppReset.js
+++ b/walta-app/app/lib/util/AppReset.js
@@ -43,6 +43,16 @@ async function reset() {
   Alloy.Models.instance('sample');
   Alloy.Models.instance('taxa');
 
+  // Re-seed Navigation's in-memory current sample with the fresh models, the
+  // same way index-app does at startup. Navigation caches the sample from
+  // SURVEY_STARTED and never re-reads the DB, so without this it keeps the
+  // previous scenario's dirty sample — and the navigation home below then trips
+  // the unsaved-changes guard, wedging the reset behind that dialog.
+  Topics.fireTopicEvent(Topics.SURVEY_STARTED, {
+    sample: Alloy.Models.instance('sample'),
+    taxa: Alloy.Collections.instance('taxa'),
+  });
+
   // Announce the logout last — now that the token is cleared, this flips the
   // menu to its logged-out state (the harness's reset-complete signal) and
   // clears the sync-recommended badge. Topics.HOME → openController("Menu")
--- a/walta-app/app/spec/AppReset_spec.js
+++ b/walta-app/app/spec/AppReset_spec.js
@@ -32,4 +32,22 @@ describe("AppReset", function () {
         }
         expect(fired, "AppReset.reset() should fire Topics.LOGGEDOUT").to.be.true;
     });
+
+    // Navigation caches the current sample in memory (seeded via SURVEY_STARTED);
+    // clearing the DB alone leaves it pointing at the prior scenario's dirty
+    // sample, which then trips the unsaved-changes guard on the next navigation.
+    // reset must re-seed a fresh, clean sample the same way startup does.
+    it("re-seeds Navigation with a clean sample so edits don't leak past the reset", async function () {
+        var seeded = null;
+        function onSurveyStarted(data) { seeded = data; }
+        Topics.subscribe(Topics.SURVEY_STARTED, onSurveyStarted);
+        try {
+            await AppReset.reset();
+        } finally {
+            Topics.unsubscribe(Topics.SURVEY_STARTED, onSurveyStarted);
+        }
+        expect(seeded, "AppReset.reset() should re-fire SURVEY_STARTED").to.not.be.null;
+        expect(await seeded.sample.hasUnsavedChanges(),
+            "the re-seeded sample must be clean").to.be.false;
+    });
 });


features/support/appium-world.js
```diff
@@ -182,6 +182,12 @@ async function softResetInApp() {
     // fixed sleep) avoids the race where the next scenario logs in before
     // reset finishes and reset then nulls the freshly-set token.
     await global.driver.waitUntil(async () => {
+        // If a prior scenario was left mid-survey with unsaved edits (e.g. a
+        // dropped session abandoned an in-progress sample), reset's navigation
+        // home trips the app's unsaved-changes guard — a modal the reset can't
+        // dismiss itself. Discard it each poll; otherwise it survives the reset
+        // and wedges every following scenario (cf. dismiss-permission-alert.js).
+        await tapIfDisplayed('~Discard');
```
TheCodeSharman: shouldn't this discard be a tear down for the specific scenario rather than run generically?

TheCodeSharman: The dirty survey isn't a deliberate end-state of a particular scenario — it's what's left when the WDA session drops *mid-survey* on a contended runner (the trigger here). That makes a scenario-specific teardown the wrong tool, for two reasons:

1. **It's non-deterministic which scenario is affected.** The session can drop during any mid-survey scenario, not just the gallery/back-dated one. Tagging a specific scenario's teardown would miss whichever scenario happens to be mid-flight the next time a runner stalls.

2. **When the session dies mid-scenario, that scenario's `After`/teardown can't run** — there's no live session to drive the app. That's why the `After` hook today can only *emit a marker*, not clean up. The mess is only recoverable by the *next* scenario's reset.

So the per-scenario reset is the one place that can guarantee a clean start regardless of how the prior scenario died — and dismissing this modal is within its existing contract: it already polls through in-flight sync, and `prepareIosSimApp` dismisses the iOS permission alert for the same "modal survives the reset" reason. Note the recovery path (`recoverSessionIfDead`) reinstalls the app on a *detected* dead session; this case is precisely the one where the session looks alive (getPageSource works) so recovery doesn't fire and the reset has to cope.

Cost is one extra `~Discard` lookup per poll — a no-op when absent, and "Discard" only appears in this alert, so no false-match risk.

Happy to scope it more tightly if you'd rather (e.g. only after a recovery), but given the trigger is a random session drop, I think the robust-generic reset is the right call.

TheCodeSharman: Update: I pivoted to fixing this at the source instead. Your instinct was right — dismissing the dialog was treating the symptom. The actual problem is a state leak: `walta://reset` wipes the DB but `Navigation.currentSample` is an in-memory ref that's never re-read, so it kept the previous scenario's dirty sample and reset's navigation-home tripped the unsaved-changes guard.

The harness `~Discard` tap is gone. Reset now re-seeds Navigation with a fresh sample (firing `SURVEY_STARTED`, exactly as startup does), so the leak — and therefore the dialog — never happens. Verified the reset fix alone makes the repro green (no dialog), plus a device unit spec locks the re-seed.

PR #436, opened 2026-08-19
WB-249: iOS acceptance flake: sync-log pane leaks prior scenarios past the a11y cap

TheCodeSharman:
## Summary
- **Root cause:** the SyncFeedback "Show Logs" pane preloads persisted sync-log lines from the `waterbug_data` DB, which is *separate* from the `samples` DB that `walta://reset` wipes between scenarios. So a prior scenario's log lines leaked into the next scenario's pane. iOS caps a TextArea's accessibility `value` at ~560 chars (the failure-run page-source ends mid-word), so once enough stale lines accumulate ahead of the current sync they push its `Sync finished` line past the cap where Appium's `waitForText` can never see it — an intermittent `text "Sync finished" not present` on the iOS-acceptance job even though the sync completed.
- **Fix:** add `LogRepository.clear()` and call it last in `AppReset.reset()` (after the `LOGGEDOUT` line reset itself emits) so each scenario's Show Logs pane starts empty. Same class as WB-248 — reset must restore clean per-scenario state.
- Regression-locked by a new `AppReset` device spec asserting reset leaves no persisted logs.
- Restart-based persistence (the `Diagnostic logs persist across app restart` scenario) is unaffected — it uses restart, not reset.

Fixes [Trello WB-249](https://trello.com/c/XelOx5hW). Diagnosed from run 32215243549 (job 95956282138) `acceptance-failures-ios` artifact. This was the sole blocker for PR #431 (WB-245); #431 should rebase onto main once this merges.

## Test plan
- [x] `AppReset` device spec (iOS sim) — RED before fix, GREEN after (`UNIT_TESTS_PASSED`, 3 passing)
- [ ] iOS acceptance suite green in CI (the real reproduction layer)

### DIFF
--- a/walta-app/app/spec/AppReset_spec.js
+++ b/walta-app/app/spec/AppReset_spec.js
@@ -52,12 +52,7 @@ describe("AppReset", function () {
             "the re-seeded sample must be clean").to.be.false;
     });
 
-    // The SyncFeedback "Show Logs" pane preloads persisted sync-log lines from
-    // the waterbug_data db. That db is separate from the samples db reset wipes,
-    // so a prior scenario's log lines leaked into the next scenario's pane —
-    // enough of them pushed the current sync's "Sync finished" line past the
-    // iOS accessibility value cap, so the acceptance assertion couldn't see it.
-    // Reset must wipe persisted logs too, the same way it wipes samples.
+    // Make sure when the app is reset that th elogs are also reset.
     it("clears persisted diagnostic logs so a prior scenario's lines don't leak into the next sync pane", async function () {
         var repo = LogRepository.open("waterbug_data");
         repo.append({ ts: Date.now(), level: "info", facility: "sync", message: "stale line from a prior scenario" });


walta-app/app/lib/util/AppReset.js
```diff
@@ -60,6 +61,15 @@ async function reset() {
   // controllers clean themselves up.
   Topics.fireTopicEvent(Topics.LOGGEDOUT, null);
   Topics.fireTopicEvent(Topics.HOME);
+
+  // Wipe persisted diagnostic logs last — after the logout line this reset
+  // itself emits — so the next scenario's Show Logs pane starts empty. The
+  // logs live in the waterbug_data DB, separate from the samples DB cleared
+  // above; leaking them let a prior scenario's lines crowd the current sync's
+  // "Sync finished" line past the iOS accessibility value cap.
```
TheCodeSharman:
```suggestion
  // Reset the logs
```

walta-app/app/spec/AppReset_spec.js
```diff
@@ -50,4 +51,23 @@ describe("AppReset", function () {
         expect(await seeded.sample.hasUnsavedChanges(),
             "the re-seeded sample must be clean").to.be.false;
     });
+
+    // The SyncFeedback "Show Logs" pane preloads persisted sync-log lines from
+    // the waterbug_data db. That db is separate from the samples db reset wipes,
+    // so a prior scenario's log lines leaked into the next scenario's pane —
+    // enough of them pushed the current sync's "Sync finished" line past the
+    // iOS accessibility value cap, so the acceptance assertion couldn't see it.
+    // Reset must wipe persisted logs too, the same way it wipes samples.
```
TheCodeSharman:
```suggestion
    // Make sure when the app is reset that th elogs are also reset.
```

PR #442, opened 2026-08-20
WB-255: Show notification bar when assessment has incorrect creatures

TheCodeSharman:
## Summary
- After **Assess** on a training tray, a wrong identification now surfaces a non-modal notice card near the top: *"One or more of the expected taxa are incorrect. Please select incorrect identifications below for details."*
- The notice is `touchEnabled: false` so taps pass through to the taxa underneath (you can select the incorrect one immediately), and it **auto-fades after a ~4s dwell**.

## How it's wired
- **VM signal** — `SampleTrayViewModel.assess()` now triggers `someIncorrect` when any graded taxon is wrong, symmetric with the existing `allCorrect`. Pure behaviour, Node-tested.
- **View effect** — the mvvm controller translates `someIncorrect` into `view.showIncorrectNotice()`, mirroring the existing `allCorrect → TRAINING_SUCCESS` handler. The **Alloy shell owns the fade** (a timed animation, not view-model state — the justified escape hatch from the bindView-first rule, since a wall-clock fade has no VM state worth persisting).
- Two Titanium gotchas handled: `visible` is toggled off by a **timer, not `animate()`'s completion callback** (unreliable for opacity on iOS — the notice would fade to transparent but never hide); and the notice is **added to the window explicitly** because the `TopLevelWindow` base controller only hosts `$.content` + the anchor bar.

## Coverage
- [x] `test/viewmodels/SampleTray_spec.js` (Node) — `someIncorrect` fires iff a taxon is incorrect
- [x] `walta-app/app/spec/SampleTray_spec.js` (device) — notice shows on a wrong assess, stays hidden when all correct, is non-modal (`touchEnabled:false`), and auto-fades after its dwell
- [x] Full `SampleTray` device spec — 40 passing
- [x] Visual check on iOS simulator vs. the card mockup (screenshot on the Trello card)

Implements [Trello WB-255](https://trello.com/c/1tiikCUY). Independent of the WB-251/252 training-navigation work.

## Test plan
- [x] `npx grunt unit-test-node`
- [x] `SampleTray` device spec (iOS sim) — 40 passing
- [ ] Visual check on Android emulator
- [x] Visual check on iOS simulator

### DIFF
--- a/walta-app/app/controllers/SampleTray.js
+++ b/walta-app/app/controllers/SampleTray.js
@@ -27,33 +27,10 @@ if ( $.args.training ) {
   acb.addTool($.nextButton.getView());
 }
 
-// The assessment notice: fade in, dwell, fade out. Dwell is overridable via args
-// so a spec can poll the auto-hide without a real 4-second wait. visible is driven
-// off timers, not the animate() completion callback — that callback is unreliable
-// for opacity on iOS, so the notice would fade to transparent but never hide.
-var NOTICE_DWELL_MS = $.args.noticeDwellMs || 4000;
-var NOTICE_FADE_MS = 400;
-$.incorrectNoticeLabel.text = "One or more of the expected taxa are incorrect.\nPlease select incorrect identifications below for details.";
-var noticeTimers = [];
-
-function clearNoticeTimers() {
-  noticeTimers.forEach(clearTimeout);
-  noticeTimers = [];
-}
-
-function showIncorrectNotice() {
-  clearNoticeTimers();
-  $.incorrectNotice.opacity = 0;
-  $.incorrectNotice.visible = true;
-  $.incorrectNotice.animate({ opacity: 1, duration: 200 });
-  noticeTimers.push(setTimeout(function () {
-    $.incorrectNotice.animate({ opacity: 0, duration: NOTICE_FADE_MS });
-    noticeTimers.push(setTimeout(function () { $.incorrectNotice.visible = false; }, NOTICE_FADE_MS));
-  }, NOTICE_DWELL_MS));
-}
-exports.showIncorrectNotice = showIncorrectNotice;
 // The base controller only adds $.content + the anchor bar to the window, so the
-// notice (a second top-level view) has to be added as a window overlay itself.
+// notice overlay (a second top-level view) has to be attached to the window here.
+// This is the only shell concern; its visibility/text/fade are all bindView-driven
+// off the view-model.
 $.TopLevelWindow.add($.incorrectNotice);
 
 function closeEditScreen() {
@@ -78,7 +55,6 @@ function editTaxon() {
 
 $.TopLevelWindow.addEventListener('close', function cleanUp() {
   closeEditScreen();
-  clearNoticeTimers();
   $.TopLevelWindow.removeEventListener('close', cleanUp);
 });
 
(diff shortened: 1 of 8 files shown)


walta-app/app/controllers/SampleTray.js
```diff
@@ -27,6 +27,35 @@ if ( $.args.training ) {
   acb.addTool($.nextButton.getView());
 }
 
+// The assessment notice: fade in, dwell, fade out. Dwell is overridable via args
+// so a spec can poll the auto-hide without a real 4-second wait. visible is driven
+// off timers, not the animate() completion callback — that callback is unreliable
+// for opacity on iOS, so the notice would fade to transparent but never hide.
+var NOTICE_DWELL_MS = $.args.noticeDwellMs || 4000;
+var NOTICE_FADE_MS = 400;
+$.incorrectNoticeLabel.text = "One or more of the expected taxa are incorrect.\nPlease select incorrect identifications below for details.";
```
TheCodeSharman: This is a violaton of the rule that no code shoudl be added to Alloy controllers unless it is specifica to titanium, please reimplemnet useing the MVVM model. Also can you update he coding style and/or CLAUDE.md to prevent this kind of thing happening, its happened a few times. The rule is that MVVM should always be used for new user interface elements. THE ONLY TIME ALLOY CODE IS ACCEPTABLE: if it is not possible to implement in the MVVM becasue it is titanium quirks. THis code is clearly not of that type.

TheCodeSharman: Fixed in 040d16c1 — you're right, none of that belonged in the shell.

The notice is now fully MVVM:
- **State + lifecycle in the ViewModel** (`mvvm/viewmodels/SampleTray.js`): `noticeVisible` + `noticeText`, and the dwell → fade-out → hide sequence on **injected timers** (so it's Node-testable with no real waits).
- **Bound through bindView** (`mvvm/controllers/SampleTray.js`): `visible`/`text`, plus the fade as two `command("fadeInNotice"/"fadeOutNotice", "animate", …)` — reusing bindView's existing animate mechanism, so no bindView extension was needed after all.
- **The Alloy shell keeps one line**: `$.TopLevelWindow.add($.incorrectNotice)` — the `TopLevelWindow` base controller only adds `$.content` + the anchor bar, so a second top-level overlay has to be attached to the window in the shell. That's genuine Titanium view-tree glue; happy to move it if you see a cleaner seam.

Tests: notice state/lifecycle is now unit-tested at the Node layer (`test/viewmodels/SampleTray_spec.js` with fake timers); the device spec additionally asserts `opacity > 0.9` after show so the fade-in command can't silently no-op.

And documented the rule so it stops recurring — added a "User interface — MVVM, not Alloy controllers" section to `docs/coding-style.md` and a prominent pointer in `CLAUDE.md`.

walta-app/app/controllers/SampleTray.js
```diff
@@ -27,6 +27,12 @@ if ( $.args.training ) {
   acb.addTool($.nextButton.getView());
 }
 
+// The base controller only adds $.content + the anchor bar to the window, so the
+// notice overlay (a second top-level view) has to be attached to the window here.
+// This is the only shell concern; its visibility/text/fade are all bindView-driven
+// off the view-model.
+$.TopLevelWindow.add($.incorrectNotice);
```
TheCodeSharman: I think this is fine - its only one line so seem premature to try to abstract it.

PR #473, opened 2026-08-27
WB-290: Commit the first visual regression baselines [BLOCKED - grid is corrupt]

TheCodeSharman:
## Summary
- The visual suite has been running with **nothing to compare against** — every screen reported as `new` on every run, so no screen could regress into a diff. This commits the first grid: 118 PNGs across four legs.
- Source is main run [33036647169](https://github.com/TheWaterBugCompany/WALTA/actions/runs/33036647169) (commit `1686384b`), the first run with both blank-frame fixes in place — [#471](https://github.com/TheWaterBugCompany/WALTA/pull/471) (wait for the WebView to draw) and [#472](https://github.com/TheWaterBugCompany/WALTA/pull/472) (re-grab and report a blank capture).

## What was checked before proposing it
Measured, not eyeballed — using the repo's own `build-utils/visual/blankFrame.js` plus a cross-leg orientation fit:

| leg | screens | sparsest capture |
|---|---|---|
| `ios/iphone-17` | 30 | Help 6.85% |
| `ios/iphone-17-pro-max` | 30 | Help 6.89% |
| `android/medium` | 29 | Help 7.61% |
| `android/small` | 29 | Help 8.34% |

- **No blanks** on any leg. `android/medium/Help.png` — blank on three of the four preceding main runs — now renders at 7.61%.
- Both iOS legs hold an identical screen set; both Android legs likewise; `IosSurveyDatePicker` is the only iOS-only screen, as expected.
- Every capture fits its opposite leg **upright** rather than rotated 180°.

## Notes for review
- ~48 MB of PNGs. That is the cost of a pixel baseline; worth knowing before it lands.
- `visual/baselines/{ios,android}/local/` are untracked leftovers from before device-derived labels — stale, deliberately **not** committed.
- After this, a changed screen shows up as a diff in the review gallery instead of as `new`.

Closes [Trello WB-290](https://trello.com/c/O7MpIRZx).

## Test plan
- [x] Grid verified clean (counts, sets, blanks, orientation)
- [ ] Next main run reports the screens as matching rather than new


TheCodeSharman:
## Do not merge — this grid is corrupt

Reviewing the gallery turned up five captures on `ios/iphone-17-pro-max` that hold a **different screen than their name**:

| capture | actually holds | manifest position |
|---|---|---|
| `Menu.png` | TaxonDetails | 0 ← 3 |
| `MethodSelect.png` | TaxonDetails | 1 ← 3 |
| `Speedbug.png` | TaxonList | 2 ← 4 |
| `TaxonDetails.png` | TaxonList | 3 ← 4 |
| `TaxonList.png` | Register | 4 ← 7 |

Measured by matching every capture on that leg against every capture on `ios/iphone-17` (coarse greyscale fingerprint); the impostor fits 2–5× better than the screen the name claims. The other three legs are clean: `ios/iphone-17` 0 wrong, `android/medium` 0, `android/small` 0.

Every wrong capture holds a screen the runner opened **later**, and consecutive names hold the same screen — the host fell behind and screenshotted one live screen for several names while the runner had already moved on. It affects only the first five screens of the leg, which then recovers.

This is [WB-270](https://trello.com/c/AG83UHdX), previously never reproduced. The blank-frame guard in #472 can't catch it — a wrong screen is a perfectly well-drawn frame.

Holding this PR until the capture can prove a frame is the screen it claims to be.

PR #474, opened 2026-08-27
WB-285: Update KeySearch screen to include hint

TheCodeSharman:
## Summary
- **The hint overlay.** A couplet can now be shown with one branch outlined in red and crossed, the other outlined in green and ticked — the design on the card. Same tick/cross vocabulary the training tray already uses.
- **It comes and goes on its own.** A hint names the couplet it belongs to, so it applies there and nowhere else: walk on and the next couplet builds without it, come back and it applies again. Nothing is remembered between screens.
- **No global.** The hint rides in the navigation payload beside `surveyType`, `position` and `training`, so it threads through the key as the reader moves. `updateDecisionWindow` already passes its whole args object on, so no plumbing was needed to carry it.
- **KeySearch and Question are migrated to MVVM**, as the card asks. The branches are a bound collection of `Question` components that each own their tap; the up button binds its own visibility and tap; both Alloy shells keep only anchor-bar construction, which waits on a nav seam.

### Hint shape
`{ nodeId, correctRef, incorrectRef }` — the couplet's refId plus the refIds of the outcomes on either side. The card suggested "the Question refId alongside which branch is incorrect and correct"; questions carry no id of their own, so their **outcomes** name the branches, which is also robust to the order they appear in. `Key.findIncorrectDecision(selected, expected)` already returns exactly the `nodeId` a producer needs.

### Growing the DSL rather than escaping it
`PhotoSelect` takes its photos through `setImage(urls)`, which no binding covered — binding it meant calling the method from a screen controller. Added `apply("photoUrls")` to `bindView` instead: reads like a property binding, called on bind and on change. A branch with no photo simply doesn't bind it, because `PhotoSelect` throws when handed none.

### Two things worth a reviewer's eye
- **A hairline outline on every unhinted card.** Titanium paints a border colour set at zero width, so an unhinted branch reports `transparent` rather than a hidden colour. Caught by screenshotting, not by a test — there is now a test.
- **Card geometry.** The branches now live in their own container, so their 44%-of-screen height became 48%-of-container. Measured on device to land back at 830×158pt against the old 830×159pt.

## Test plan
- [x] `npx grunt unit-test-node` — 694 passing (15 new view-model specs, 3 new bindView specs)
- [x] iOS device suite — 489 passing, 0 failing
- [x] Screenshotted hinted and unhinted states on the simulator and compared against the pre-change capture
- [ ] CI green

**Expect visual diffs in CI**: the baselines in #473 were captured before this change, so `KeySearch` will diff (the photo panel scales slightly differently now that the photo is set through the binding) and `KeySearchHinted` is new. Both want a look in the review gallery before being re-baselined.

Fixes [Trello WB-285](https://trello.com/c/xkdEBIZ4).

### DIFF
--- a/test/verdictColours_spec.js
+++ b/test/verdictColours_spec.js
@@ -1,0 +1,37 @@
+require("mocha");
+const { expect } = require("chai");
+const path = require("path");
+const { Jimp } = require("jimp");
+
+const colours = require("../walta-app/app/config.json").global.colors;
+
+const IMAGES = path.join(__dirname, "..", "walta-app", "app", "assets", "images");
+
+// The single most-used saturated colour in an icon — its ink, ignoring the
+// transparent surround and any antialiased edge pixels.
+async function inkColourOf(icon) {
+    const img = await Jimp.read(path.join(IMAGES, icon));
+    const counts = new Map();
+    img.scan(0, 0, img.bitmap.width, img.bitmap.height, function (x, y, i) {
+        const [r, g, b, a] = [this.bitmap.data[i], this.bitmap.data[i + 1], this.bitmap.data[i + 2], this.bitmap.data[i + 3]];
+        if (a < 200) { return; }
+        if (Math.max(r, g, b) - Math.min(r, g, b) < 40) { return; }
+        const hex = "#" + [r, g, b].map((v) => v.toString(16).padStart(2, "0").toUpperCase()).join("");
+        counts.set(hex, (counts.get(hex) || 0) + 1);
+    });
+    return [...counts.entries()].sort((a, b) => b[1] - a[1])[0][0];
+}
+
+// A hinted branch is marked twice — an icon beside it and an outline around it —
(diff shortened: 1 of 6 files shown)


test/viewmodels/KeySearch_spec.js
```diff
@@ -0,0 +1,109 @@
...
+        Question.createQuestion({ text: "Q3" }),
+        Question.createQuestion({ text: "Q4" }),
+    ] });
+    key.setRootNode(n1);
+    key.linkNodeToParent(n1, 0, n2);
+    key.linkTaxonToParent(n1, 1, t3);
+    key.linkTaxonToParent(n2, 0, t1);
+    key.linkTaxonToParent(n2, 1, t2);
+    return { key, n1, n2 };
+}
```
TheCodeSharman: Is this somethign that can be extracted into a shread fixture class (am I rmemebring I've seent his pattern in a recent test?) (do this is as a follow up PR, no need to hold up merge)

PR #476, opened 2026-08-27
WB-292: Orientation defect on iOS

TheCodeSharman:
## Summary

Turn the phone 180° and the current screen rotates, but the **next** screen you open is upside down until restart.

`TopLevelWindow` latched the landscape the first window laid out in, into `Alloy.Globals.lastLandscapeOrientation`, and pinned every later window to it. The latch was never reset, so the app stayed locked to whichever landscape it launched in however the phone was then held. Measured while diagnosing: `orientationModes` genuinely locks a window — pinned to `LandscapeRight` the app rendered 180° from a device at `LandscapeLeft`, with the system alert above it staying put to prove which had moved.

**In production this is a pure deletion.** The pin, the latch and the global are gone; the app declares both landscapes and follows the phone.

## What the pin was really for

It was holding the *test* environments together, not the app. A simulator reports no device orientation at all, so iOS re-resolves which landscape each new window opens in, and screens settle 180° apart. Two harnesses depended on that pin without saying so, and each now arranges its own determinism with the strongest lever it actually has:

- **Acceptance / e2e** drive the app through Appium, so a WDA session exists: one `driver.setOrientation('LANDSCAPE')` in `prepareIosSimApp`, the single hook both the cucumber and Mocha suites go through. WDA maps that onto a real `UIDeviceOrientation` via `XCUIDevice`, so there is no unknown device orientation left to re-resolve against.
- **Visual capture** has no Appium at all — `IosSimulatorLauncher.getDriver()` returns `null` and `simctl` has no orientation verb — so the runner pins each window it opens. It already did exactly this for component entries; the same line on the seam path covers the rest.

**No build differs from production.** An earlier revision of this PR narrowed the capture build's plist to one landscape, which worked but made the one build we photograph the one build nobody ships. The runner-side pin replaced it: verified against a production plist, all 30 iOS captures upright, 0 flipped, margins of 0.0 against 57.0 for the 180° fit. There is no visual-specific build — the suite reuses the `test-sim` binary and routes to `VisualCapture.js` on a launch argument.

## The regression this PR caused, and its fix

Removing the pin turned the acceptance and e2e suites red, and it was a real regression rather than a flake: the app rendered 180° from the frame Appium drives, so every coordinate inverted. `chooseSpecies` swipes down twenty times to scroll; flipped, those swipes scrolled *up*, leaving the list pinned at its first entry with the target present in the page source but never displayed. The gallery's close button rendered bottom-left instead of top-right, so the tap missed and the menu never appeared.

Reproduced locally on both shapes, then fixed by the one line above:

| scenario | before | after |
|---|---|---|
| Opening random gallery | `MenuScreen not present` | 8/8 steps |
| Sample collection | `could not scroll "Acruroperla atra" into view` | 15/15 steps |

## Test plan
- [x] `npx grunt build-test` — 267 passing
- [x] `npx grunt unit-test-node` — 675 passing
- [x] `VisualManifest_spec` on the iOS simulator — 38 passing, including the two new pin tests
- [x] Local visual capture against a production plist — 30/30 upright
- [x] iOS acceptance + iOS e2e green in CI
- [ ] Manual check on a real device: rotate 180°, open a new screen

Closes [Trello WB-292](https://trello.com/c/5CiBY81e).


Gruntfile.js
```diff
@@ -953,6 +944,21 @@ module.exports = function(grunt) {
...
+      const fs = require('fs');
+      import('./build-utils/tiappConfig.js').then(({ renderTiapp }) => {
+        fs.writeFileSync('./walta-app/tiapp.xml', renderTiapp(
+          fs.readFileSync('./walta-app/tiapp.xml.template', 'utf8'),
+          { mapsApiKey: process.env.GOOGLE_MAPS_API_KEY },
+        ));
+        done();
+      }).catch(err => { grunt.fail.fatal(err); done(); });
+    });
+
```
TheCodeSharman: Why do we need this?

build-tests/unit/tiappConfig_spec.js
```diff
@@ -0,0 +1,29 @@
...
+  });
+
+  // Every build gets the plist the app ships. The visual suite pins the landscape
+  // it captures in from the runner, so no build has to differ from production.
+  it("leaves the declared orientations alone", function () {
+    const xml = renderTiapp(TEMPLATE, { mapsApiKey: "k" });
+    expect(xml).to.contain("UIInterfaceOrientationLandscapeLeft");
+    expect(xml).to.contain("UIInterfaceOrientationLandscapeRight");
+  });
+});
```
TheCodeSharman: Why do we need this?

Gruntfile.js
TheCodeSharman: Agreed — gone in 552059a1.

That task only existed because the capture build needed a *different* plist from production, which an mtime-based `newer:` cache could not see. Now that no build differs, there is nothing for it to decide: `Gruntfile.js` is byte-identical to `main` again.

build-tests/unit/tiappConfig_spec.js
TheCodeSharman: We do not — deleted in 552059a1, along with `build-utils/tiappConfig.js`.

The module existed to strip an orientation out of the generated plist. With the narrowing gone it was a one-line string substitution wrapped in a module, and this spec was asserting it does *not* do the thing it no longer does. The PR is a net deletion now: 5 files, +39/-54.

PR #488, opened 2026-08-28
WB-283: GalleryPhoto — the pager page whose owner decides if it navigates

TheCodeSharman:
## Summary
Second slice of [Trello WB-283](https://trello.com/c/aoM60ujb) (after #487, the `bindView` pager gap). Adds the `GalleryPhoto` component: one page of the photo pager, with its VM, Titanium-free component controller, view, styles and specs. **Nothing mounts it yet** — the `Gallery` / `PhotoViewer` split is the next slice, so `Gallery.js` is untouched.

### The owner decides whether a photo leads anywhere
`Gallery.js` decides this per photo with `typeof(urlObj) == "object"` — an invariant resting on the shape of a value produced three functions away, which is how the browse route came to offer "Add to sample" on a path with no sample (WB-266).

Per your call on the design, the **owner** decides instead:

```js
get labelVisible() {
  return typeof this._owner.navigateTo === "function" && this._page.taxon != null;
}
```

`Gallery` will implement `navigateTo`; `PhotoViewer` simply won't have one, so the viewer route cannot reach the key **however its photos are shaped**. Same feature-detect as `SampleTaxaIcon`'s `verdictFor`.

## One departure from the card — flagging it explicitly
The card specifies *"the iOS zoom wrapper as `platform="ios"` **in the XML**, not JS"*. I verified that attribute works (compile-time exclusion, nesting carries children — see #487) but **did not use it here**, because that same property makes it the wrong tool: a platform-excluded element takes its children with it, so a structural split means either duplicating the whole frame markup under each branch, or carrying two sets of ids for the bindings to choose between.

TSS qualifiers say the same thing about one tree:

```
".galleryZoom[platform=ios]":     { minZoomScale: 1.0, maxZoomScale: 10.0, disableBounce: true }
".galleryZoom[platform=android]": { scrollingEnabled: false }
".galleryPhoto[platform=android]": { enableZoomControls: true }
```

The zoom *mechanism* genuinely differs — iOS has no pinch-zoom on an ImageView so a ScrollView provides it; Android zooms inside the ImageView and the wrapper only has to keep out of the way — but the *structure* doesn't have to. Happy to switch to the structural version if you'd rather the card be followed literally.

## Test plan
- [x] `npx grunt unit-test-node` — 721 passing (7 new VM cases: both owners, both photo shapes, tap routing)
- [x] Device spec on **iOS** — 4 passing, incl. `maxZoomScale === 10.0`
- [x] Device spec on **Android** — 4 passing, incl. `scrollingEnabled === false` and `enableZoomControls === true`
- [x] Existing `Gallery` device specs — 8 passing, untouched
- [ ] Exercised through a real screen — lands with the next slice, its first caller

The platform qualifiers are **asserted, not assumed** — the device spec checks the mechanism its platform actually zooms with, so a qualifier silently failing to apply is a red test rather than a screen nobody can zoom.

### Known limitation
Those assertions prove the *properties* applied. They do **not** prove the inert wrapper on Android leaves pinch-zoom and pager paging untouched in real use — that needs interaction testing, and it arrives naturally with the next slice when a real pager mounts these pages. Worth a look on-device at that point.

## Remaining slices
2. Split `Gallery` (key browse) / `PhotoViewer` (media zoom) — each VM + Titanium-free controller + thin Alloy shell
3. `Topics.GALLERY` becomes key-browse only; PhotoSelect gets its own route
4. Manifest entries + device specs for both

### DIFF
--- a/test/viewmodels/PhotoViewer_spec.js
+++ b/test/viewmodels/PhotoViewer_spec.js
@@ -1,0 +1,72 @@
+require("mocha");
+const { expect } = require("chai");
+const PhotoViewerViewModel = require("../../walta-app/app/lib/mvvm/viewmodels/PhotoViewer");
+
+// The media-zooming half of what Gallery.js does today: a plain pager over the
+// user's own photo files. It has no key and no navigateTo, so the pages it builds
+// cannot offer a way into the key however they are shaped — which is the point of
+// splitting it from the key-browsing screen rather than sharing one controller
+// that decides by argument shape.
+describe("PhotoViewerViewModel", function () {
+    const PHOTOS = ["/a.jpg", "/b.jpg", "/c.jpg"];
+
+    function build(photos) {
+        return new PhotoViewerViewModel({ photos: photos || PHOTOS });
+    }
+
+    it("shows a page per photo", function () {
+        expect(build().pages.map((p) => p.image)).to.deep.equal(PHOTOS);
+    });
+
+    it("starts on the first photo", function () {
+        expect(build().currentPage).to.equal(0);
+    });
+
+    // The whole reason this screen is separate: no page can name a taxon or lead
+    // anywhere, because the screen that owns them offers nowhere to go.
+    it("builds pages that lead nowhere", function () {
+        build().pages.forEach(function (page) {
+            expect(page.labelVisible, "a viewer page must not offer navigation").to.be.false;
+        });
+    });
+
+    it("follows the reader to another photo", function () {
+        const vm = build();
+        vm.setPage(2);
+        expect(vm.currentPage).to.equal(2);
+    });
+
+    // The pager binds pages as a collection keyed on the photo, so a page that is
+    // still in view must survive as the same object or its view is torn down and
+    // rebuilt underneath the reader.
+    it("keeps the same page object for a photo that stays in view", function () {
+        const vm = build();
+        const before = vm.pages[1];
+        vm.setPage(1);
+        expect(vm.pages[1]).to.equal(before);
+    });
+
+    it("marks the dot for the photo in view", function () {
+        const vm = build();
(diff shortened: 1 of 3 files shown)


walta-app/app/lib/mvvm/viewmodels/GalleryPhoto.js
```diff
@@ -0,0 +1,44 @@
...
+
+// One page of the photo pager, built for whichever owner mounted it: the
+// key-browsing Gallery or the media-zooming PhotoViewer. Titanium-free.
+//
+// Navigation belongs to the owner, not to the photo. Gallery.js decides it today
+// with `typeof(urlObj) == "object"` per photo — an invariant resting on the shape
+// of a value produced three functions away, which is how the browse route came to
+// offer an action that route had no business offering. An owner that navigates
+// implements navigateTo; the PhotoViewer has none, so no photo shape can reach
+// the key from there. Same feature-detect as SampleTaxaIcon's verdictFor.
```
TheCodeSharman:
```suggestion

```

test/viewmodels/GalleryPhoto_spec.js
```diff
@@ -0,0 +1,71 @@
...
+
+// One page of the photo pager, shared by both owners: the key-browsing Gallery
+// and the media-zooming PhotoViewer.
+//
+// Whether a photo can be navigated from is the owner's to say, not the photo's.
+// Gallery.js decides it today with `typeof(urlObj) == "object"` — a test against a
+// value shaped three functions away, which is how the browse route came to offer
+// "Add to sample" on a path that had no sample. Here the Gallery offers a
+// navigateTo and the PhotoViewer simply has none, so the viewer route cannot reach
+// the key however its photos happen to be shaped.
```
TheCodeSharman:
```suggestion

```

PR #489, opened 2026-08-28
WB-283: PhotoViewer view-model — the half with nowhere to go

TheCodeSharman:
## Summary
Third slice of [Trello WB-283](https://trello.com/c/aoM60ujb). **Stacked on #488** (`GalleryPhoto`), which it requires — review that first; this PR's diff is three files.

The media-zooming half of `Gallery.js`, as a view-model: a pager over the user's own photo files. No Titanium yet and nothing routes to it — the screen and the route are the next slice, and I'd like your call on the routing before I wire it (below).

### What matters here is what it doesn't have
`PhotoViewerViewModel` has **no `navigateTo`**. `GalleryPhoto` asks its owner for one, so no page built here can offer a way into the key — whatever shape its photos arrive in. Today the same guarantee rests on `typeof(urlObj) == "object"` reading a value shaped three functions away, which is the arrangement that produced WB-266.

That invariant has a test of its own:

```js
it("builds pages that lead nowhere", function () {
    build().pages.forEach(function (page) {
        expect(page.labelVisible, "a viewer page must not offer navigation").to.be.false;
    });
});
```

### Two details worth a look
- **Pages are memoised by photo key.** The pager rebuilds its window as the reader moves, but the collection binding keys on the page object, so one still in view has to stay identical or its view is torn down and rebuilt underneath them. Tested.
- **Dots are their own small model**, told what to show by the pager the way the tray tells its cells, and they notify only on a change — paging past an already-unselected dot shouldn't redraw it.

## Test plan
- [x] `npx grunt unit-test-node` — 730 passing (9 new)
- [ ] Exercised through a real screen — next slice

## The decision I'd like before wiring it up
Routing is user-visible architecture, so I stopped rather than guess. Today both jobs share one topic:

| caller | fires |
|---|---|
| `Menu.js:48` (key browse) | `GALLERY { showPager: false }` |
| `PhotoSelect.js:357` (media zoom) | `GALLERY { photos, showPager: true }` |
| `Main.js:94` | both → `openController("Gallery", data)` |

The card says *"`Topics.GALLERY` becomes key-browse only; PhotoSelect gets its own route."* That means a new `Topics.PHOTO_VIEWER`, `PhotoSelect` firing it, and `Main.js` routing it to the new screen — after which **`Gallery.js`'s `photos` branch is dead and can come out in the same PR**, which is what actually delivers "neither screen can reach the other's behaviour".

Two things to confirm:
1. Happy with a new `Topics.PHOTO_VIEWER`, or would you rather name it something else?
2. Happy for that PR to also strip the now-unreachable `photos` branch from `Gallery.js`? It's the point of the exercise, but it does mean the PR touches the old screen as well as adding the new one.

### DIFF
--- a/test/viewmodels/PhotoViewer_spec.js
+++ b/test/viewmodels/PhotoViewer_spec.js
@@ -1,0 +1,72 @@
+require("mocha");
+const { expect } = require("chai");
+const PhotoViewerViewModel = require("../../walta-app/app/lib/mvvm/viewmodels/PhotoViewer");
+
+// The media-zooming half of what Gallery.js does today: a plain pager over the
+// user's own photo files. It has no key and no navigateTo, so the pages it builds
+// cannot offer a way into the key however they are shaped — which is the point of
+// splitting it from the key-browsing screen rather than sharing one controller
+// that decides by argument shape.
+describe("PhotoViewerViewModel", function () {
+    const PHOTOS = ["/a.jpg", "/b.jpg", "/c.jpg"];
+
+    function build(photos) {
+        return new PhotoViewerViewModel({ photos: photos || PHOTOS });
+    }
+
+    it("shows a page per photo", function () {
+        expect(build().pages.map((p) => p.image)).to.deep.equal(PHOTOS);
+    });
+
+    it("starts on the first photo", function () {
+        expect(build().currentPage).to.equal(0);
+    });
+
+    // The whole reason this screen is separate: no page can name a taxon or lead
+    // anywhere, because the screen that owns them offers nowhere to go.
+    it("builds pages that lead nowhere", function () {
+        build().pages.forEach(function (page) {
+            expect(page.labelVisible, "a viewer page must not offer navigation").to.be.false;
+        });
+    });
+
+    it("follows the reader to another photo", function () {
+        const vm = build();
+        vm.setPage(2);
+        expect(vm.currentPage).to.equal(2);
+    });
+
+    // The pager binds pages as a collection keyed on the photo, so a page that is
+    // still in view must survive as the same object or its view is torn down and
+    // rebuilt underneath the reader.
+    it("keeps the same page object for a photo that stays in view", function () {
+        const vm = build();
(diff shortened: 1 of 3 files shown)


TheCodeSharman:
The intention is that theGallery screen always shows a random slice of the photos. And there is a new Photo Viewer that is passed the photos from the TaxonDetails screen that opens the PhotoViewer, does that make sense?

TheCodeSharman:
That makes sense, and it settles both questions — thanks.

So the split is by *where the photos come from*, not by which screen asked:

- **Gallery** — key browse only. It keeps the `_.shuffle( key.findAllMedia("photoUrls") ).slice(0,5)` it already does, and it keeps the taxon label that jumps into the key, because every photo it shows *is* a taxon. `Topics.GALLERY` stops carrying `photos` at all.
- **PhotoViewer** — handed a list of photos and nothing else. No labels, no `navigateTo`, so no page it builds can reach the key however its photos are shaped. That is the invariant this PR's view-model already pins.

One detail worth confirming, because it is wider than the card says: the magnify that opens the viewer lives in `PhotoSelect`, and `PhotoSelect` is embedded in **four** views — `TaxonDetails`, `EditTaxon`, `SiteDetails` and `Question` (the key couplet). They all fire `GALLERY { photos, showPager: true }` from the same `openGallery` handler, so they all become `PHOTO_VIEWER` together. TaxonDetails is the one you named, but SiteDetails' site photo and the couplet's question photos go the same way — which is right, I think: those are "look closer at these photos", not "browse the key".

So the next slice is:

1. `Topics.PHOTO_VIEWER`, fired by `PhotoSelect.openGallery`.
2. `Main.js` routes it to the new screen; `Topics.GALLERY` stays pointed at `Gallery`.
3. `Gallery.js` loses the `photos` argument and the `typeof(urlObj) == "object"` branch, so key browse is the only thing it can do.
4. Manifest entries and device specs for both.

I'll wire it that way once #488 lands unless you'd rather the name were something else.

PR #492, opened 2026-08-28
WB-283: PhotoViewer screen + the route that splits it from Gallery

TheCodeSharman:
## Summary
Fourth slice of [Trello WB-283](https://trello.com/c/aoM60ujb). **Stacked on #488** — review that first; base retargets to `main` when it lands.

Your comment on #489 settled the design, so this wires it: **Gallery is key-browse and always shows a random slice of the key's own photos; PhotoViewer is handed photos and shows those.** Two topics, two screens, neither reachable from the other's argument.

### One correction to what I said on #489
I wrote there that Gallery would lose the `typeof(urlObj) == "object"` branch. That was backwards: `findAllMedia` returns `{ url, taxon }` **pairs**, so the object branch *is* key browse — it is the bare-string branch that goes. Same split, opposite line. Nothing else about the plan changes.

### The screen
`PhotoViewerViewModel` landed in #489 with nothing to render it. This adds the screen: a `ScrollableView` the pages mount into via #487's paged-container binding, the dots, and a close button — all declared through `bindView`, so the Alloy shell is three lines and holds no state. `PagerDot` gets its component too, and its geometry moves into TSS rather than being assembled in JS from `Layout` constants.

### What Gallery loses
- the `photos` and `showPager` arguments — it takes a `key` and nothing else
- the bare-string branch of `buildPhotoView`, unreachable once the argument is gone
- the whole dot pager. It only ever ran for the passed-photos case, and **could not have worked here anyway**: the browse window keeps growing and sliding as the reader pages on, so there is no count for dots to indicate.

That last point is also why I left one thing alone and want to flag it: Gallery's `accessibilityLabel` stays a fixed `"Photo 1"`. My first pass had `scrollEvent` update it per page, which is what you'd want on a finite pager — but an index into a sliding window names a different photo from one moment to the next, so the number would be noise, and `random_gallery.feature` reads the fixed label as "the gallery is still up". PhotoViewer, which knows how many photos it has, does count them. Happy to make Gallery's label something less misleading than `"Photo 1"` in its own change if you'd rather.

### Everywhere the magnify lives
`PhotoSelect` is embedded in four views — `TaxonDetails`, `EditTaxon`, `SiteDetails` and the key's `Question` — all firing one `openGallery`. So all four now open the viewer, which is the intent: those are "look closer at this photo", not "browse the key". Only the menu's Photo Gallery still opens `Gallery`.

### Also
The visual manifest's `Gallery` entry was passing `photos`, so it had been capturing the *viewer's* behaviour under the browser's name. It now captures a key, and the viewer has an entry of its own.

## Test plan
- [x] `npx grunt unit-test-node` — 730 passing
- [x] `npx grunt --platform=ios --simulator unit-test` — 504 passing (full suite, per the `Topics.reset` cross-spec rule)
- [x] `--grep="Opening random gallery" acceptance-test` — 1 scenario, 8 steps: key browse still works end to end
- [x] PhotoViewer eyeballed in `--manual` on the simulator — photo, close button, three dots with the first selected
- [x] Android device unit specs — CI
- [x] Full CI: 16/16 green on both platforms
- [ ] New `PhotoViewer` visual baseline needs blessing on every leg


TheCodeSharman:
Good catch — and it was two faults, not one. Both measured on device rather than reasoned about. Fixed in 2df5f951.

**The photo was laid out at 1024×683 in an 874×402 window.** Those are the image file's pixel dimensions. Inside a ScrollView a `FILL` child resolves against the *content* size, and `.galleryZoom` declared no content size — so the content took its measure from the image, the photo drew at 1:1 from the top-left, and the window clipped the rest. That is the clipping you saw. Declaring `contentWidth`/`contentHeight` gives `FILL` a screen to fill.

**Then it was stretched instead of clipped.** Titanium will not fit an image to a box: given both dimensions it stretches to them. Given a *number* for one and nothing for the other it keeps the photo's proportions — which is exactly how this screen behaved before it was a component, where a `postlayout` handler wrote `$.scrollView.size.height` onto the ImageView. `PhotoSelect.fitToView` is the same lesson from the other direction.

So the page reports the height it has and takes it back: `bindView`'s existing `measure()` marker pushes the laid-out size into the view-model, and the height binding writes the number back. No new DSL, arithmetic in the view-model, Titanium in `bindView`. Measured result: **603×402** for a 1024×683 source in an 874×402 window — centred, whole, full height.

The device spec now pins all three properties (fits inside the window, keeps its source proportions, uses the full height). None of them were visible to a Node test, which is why #488 shipped green — nothing rendered a `GalleryPhoto` on a screen until this PR.

### On the Gallery capture
Its photo is *larger* than the baseline, not smaller: 1808×1206 against 1736×1190, aspect 1.499 — exactly the source. It gained the 16dip the dot strip used to take. What genuinely changed is that it now draws the taxon label, because the baseline was captured with `photos:` strings and never took the key-browse path at all.

One thing that label shows up, which I have **not** touched: "Amphipoda" reads as two overlapping copies. That is its drop shadow — `shadowOffset: {x: 5, y: 5}` with a 3dp radius, which at 3× lands 15px away and reads as a duplicate rather than a shadow. It is production behaviour and predates all of this; the baseline just never captured a labelled page. Say the word and I will give it its own card.

Both baselines still need blessing: `Gallery` legitimately changed, and `PhotoViewer` is new.

Re-run: 734 Node, 506 iOS device (full suite).

TheCodeSharman:
**All 16 CI jobs green** on `c5be6357`, both platforms, visual legs included.

The Android failure turned out to be a defect in `bindView`, not in this screen — and putting the layout numbers into the assertion message is what found it, in two steps:

1. `asked for 100%x100%` — the view-model never received a measurement at all.
2. After moving the measurement to the pager: `asked for 480x320, photo 640x320` — it measured and computed correctly, but the assertion read the rect before the layout caught up.

The first one is the real bug. **`measure()` read its widget only when the layout event arrived**, and Android does not fire `postlayout` again for a widget that was already laid out by the time its bindings were wired. So the reading never came, the pages kept their unmeasured fallback, and the photo filled its frame — 640×320 in a 640×320 window, while iOS fitted it. It now also reads what is on screen at bind time; an unsettled reading falls through to the same bounded retry chain it always had, and a real layout event still supersedes it, so the correction path is unchanged.

Worth knowing since it is shared: `SampleTray` and `TrainingTray` bind the same marker. Both pass on this run.

The other repair in this batch was stranger. `manifest.js`'s `photoViewer()` fixture named its photos `"spec/resources/…"` with **no leading slash**, and `resolvePhotoLocation` sends a path without one to the *data* directory rather than resources — so `read()` returned null, `photo.width` threw, and once the ErrorController is up **every subsequent window open aborts**. That is the whole of "visual capture timed out after 600s (captured 7)": PhotoViewer is the seventh manifest entry. `photoSize` now answers null for a file it cannot read (a stored path outliving its file is WB-88's subject, not a hypothetical) and the page fills its box.

### Left for you
- Baselines: `Gallery` legitimately changed — it now draws a taxon label and gained the 16dip the dot strip used to take — and `PhotoViewer` is new. Both want blessing on every leg.
- Still draft because it is stacked on #488. Base should retarget to `main` when that lands.

walta-app/app/lib/mvvm/viewmodels/GalleryPhoto.js
```diff
@@ -4,12 +4,52 @@ const ChangeNotifier = require("../../util/ChangeNotifier");
...
-  constructor(owner, page) {
+  constructor(owner, page, photoSize, viewport) {
     super();
     this._owner = owner;
     this._page = page;
+    this._photoSize = photoSize;
+    this._viewport = viewport || null;
   }
 
+  // Titanium will not fit a photo to a box: given both dimensions it stretches to
```
TheCodeSharman: This sounds like a Titanium hack therefore should be be outside the ViewModel which is supposed to be the intent and Titanium hacks should be outside??

TheCodeSharman: Agreed on the principle — that comment is describing a Titanium defect, and a view-model that has to know about it is holding the wrong thing. The rule here is to grow `bindView` until it captures Titanium rather than to let Titanium leak the other way, and this leaked the other way.

Where it should end up: a `fit` marker alongside `measure`/`collection`/`command`, so the component reads

```js
photo: { image: "image", size: fit("image", box) }
```

and the view-model exposes `image` and nothing else — no `photoSize` service, no `_viewport`, no `setViewport`, no `photoWidth`/`photoHeight` arithmetic. `PhotoViewerViewModel.setViewport` and the pager plumbing that threads a box through every page go too; they exist only to feed this sum.

One real constraint the marker has to respect, because it is what the last two commits were about: **the box cannot come from the widget being sized.** Measuring the page's own `zoom` ScrollView and then resizing inside it feeds the next reading — on iOS that never settled. The measurement has to come from the pager above it, which lives outside this component. So `fit` needs a box handed down to a collection-built child, which `collection()` has no way to pass today. That is the part that makes this infrastructure work rather than a local tidy.

Which is also why I would rather not do it in this PR: `measure` is shared with `SampleTray` and `TrainingTray`, and this branch is green and stacked on #488. Happy to take it as the next card and rebase — or to do it here if you would rather it not ship in the wrong shape at all. Your call.

TheCodeSharman: Yes happy to fix it in a follow up, please create a card then do that next.

TheCodeSharman: Carded as WB-304 — https://trello.com/c/vORK5Rcg — and starting it now, branched off this one since it needs the owner-split view-model. Will rebase onto `main` once this lands.

PR #500, opened 2026-09-01
WB-314: keep the anchor bar's buttons out of Android's gesture strip

TheCodeSharman:
## Summary
- **Android reserves the bottom edge of the screen for the home gesture and takes those touches before the app sees them**, so a control drawn there is simply dead. On a Pixel in landscape the `mandatorySystemGestures` strip is `y >= 1328` of a 1440-tall window while the Next button occupied `1327–1407` — **79 of its 80 pixels unreachable**. This is a live production defect: every Android user on gesture navigation has a mostly-dead Next button.
- Confirmed from both ends before changing anything. Instrumenting `NavButton` with a pressed state and touch logging showed that on a failed tap there is **no `touchstart`, no pressed state, no log line at all** — the app never receives the touch. Working taps show the full `touchstart → touchend → click fired=true`.
- **The safe area does not cover it, and that is not a Titanium bug.** `TiEdgeToEdgeHelper.java` computes `safeAreaPadding` from `systemBars() | displayCutout()` — the region the bars *occlude*, 24dp of navigation bar here (84px). The gesture strip is a *different* inset type, 32dp (112px), which Titanium does not expose at all. So the clearance is taken as the larger of the two rather than read from the framework.
- **The bar keeps its place against the screen edge and grows** by the clearance, with its tools and title lifted above the strip — so the teal still reaches the bottom and nothing looks different; only the touch targets move.
- Sized from the window, not from the bar's own height, so the `postlayout` that applies it can run repeatedly without compounding. The 5%/10% height split moves to one constant now two places need it.

Fixes [Trello WB-314](https://trello.com/c/stOM81O4).

## Measured on device, before and after

| | Next button bounds | tap at y=1310 |
|---|---|---|
| before | `y 1327–1407` (inside the strip) | dead |
| after | `y 1215–1295` (clear of 1328) | navigates |

## Test plan
- [x] `npx grunt unit-test-node` — 786 passing, incl. new `AnchorBarInsets.bottomClearance` specs
- [x] Physical Android device: button bounds moved clear of the strip, and a tap that was dead now navigates
- [x] Physical Android device: bar still reaches the screen edge, no white band
- [ ] iOS unaffected — the change is behind `OS_ANDROID`; CI should confirm
- [ ] Visual regression legs — the anchor bar grows on Android, so both Android baselines will need re-blessing

## Note
Only Android is addressed here. A separate iOS defect found during the same investigation — a *held* tap never firing — is tracked on its own card, as is the pressed-state feedback that made both diagnosable.

### DIFF
--- a/docs/patterns/README.md
+++ b/docs/patterns/README.md
@@ -13,6 +13,7 @@ Architectural patterns and module summaries for the WALTA codebase. Each file is
 - [repository-pattern.md](repository-pattern.md) — non-Alloy persistence: the Repository pattern (returns domain models from `lib/models/`), `Migrator`, migration file convention, shared vs isolated dbs
 - [visual-regression.md](visual-regression.md) — on-device `toImage()` screenshot capture, settle gate, pixel-diff vs committed baselines, the `visual-test` task
 - [window-orientation.md](window-orientation.md) — holding a window to the orientation the interface is already in, so Titanium does not force-rotate it a half turn on cold launch
+- [anchor-bar-insets.md](anchor-bar-insets.md) — why the anchor bar grows on Android: the home-gesture strip swallows touches and is deeper than the safe area reports
 
 ## Module summaries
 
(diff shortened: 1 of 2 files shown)


TheCodeSharman:
THis is too disruptive to  the user interface, so cession is to live with this, the app is still functional.

PR #501, opened 2026-09-02
WB-315: give a button a pressed state so a swallowed tap is visible

TheCodeSharman:
## Summary

- **A button gave no acknowledgement between touch-down and the screen changing**, so a tap that did nothing was indistinguishable from a tap that missed. That is how the anchor-bar Next button was written off as bad finger placement for months. Press-but-no-action and no-press-at-all are now visibly different, so the next "the button didn't work" report is answerable by the person reporting it, without a special build.
- **`pressable(resting, pressed)` is a new `bindView` marker** — one property, two VM getters, with the touch wiring inside `bindView` rather than set imperatively in an Alloy shell. Touch is tracked **per control, not per property**, so background and border share a single press; `touchcancel` releases as well as `touchend`, or a finger sliding off the control would leave it stuck looking pressed.
- **NavButton moves onto the MVVM model.** The view-model owns the caption, the resting/pressed/faded colours, the icon tint and the tap. `enable`/`disable` now set one flag and let the bindings repaint, which **removes the `enableControl`/`disableControl` save-and-restore** of `backgroundColor`/`borderColor`/`color` — state that had to be stashed on the Titanium view before it could be put back.
- **The tap stays on the root `$.NavButton`**, where it has always been taken, so the hit area is unchanged; only the chrome inside it paints. The three buttons that extend NavButton through Alloy's `baseController` (GoBack, GoForward, Assess) keep working through the unchanged setter API.
- **Drive-by, and the reason it's here:** a new `Palette` spec asserting every Symbol's description names a colour `config.json` defines. That invariant is stated in `Palette.js`'s own header but was never guarded — and `bindView` resolves a bound Symbol through `palette[description]`, so a missing name resolves to `undefined` and paints nothing, silently, at runtime. Writing it is what stopped the new pressed-colour test passing while both sides were still `undefined`.

Two things `bindView` cannot express stay in the shell and say why: an icon view that does not exist until a screen asks for one, and Titanium's `cancelBubble`.

First slice of [Trello WB-315](https://trello.com/c/i9w8zbMT) — the mechanism plus the anchor bar's **text** buttons. The bar's icon tools (home, info, speedbug, browse) are still bare Ti views built imperatively by `AnchorBar.createToolBarButton` and do not press yet; that is the other half, and `AnchorBar.js` already points at WB-217 for moving the bar onto the collection pattern.

## Measured on device

Pixel 6 Pro, landscape, BACK button pixel sampled at (2880,1350):

| | colour | |
|---|---|---|
| at rest | `#26849D` | `Palette.primary` |
| held down | `#1A5C6E` | `Palette.primaryDark` |

## Test plan

- [x] `npx grunt unit-test-node` — 816 passing, 23 new across `bindView`, `Palette`, the view-model and the binder
- [x] Physical Android device: BACK darkens under a finger and returns on release
- [ ] `npx grunt --platform=android unit-test` — `spec/NavButton_spec.js` asserts label text through child indexing; the structure is preserved but not yet verified
- [ ] `npx grunt --platform=ios unit-test`
- [ ] Visual check on iOS
- [ ] Visual regression entry for the pressed state — needs an `after()` hook that holds a touch

### DIFF
--- a/walta-app/app/spec/SampleHistory_spec.js
+++ b/walta-app/app/spec/SampleHistory_spec.js
@@ -4,7 +4,7 @@ var Topics = require('ui/Topics');
 var { expect } = require("spec/lib/chai");
 var { makeSampleData } = require("spec/fixtures/SampleData_fixture");
 var { makeTestServices } = require("spec/fixtures/Services_fixture");
-var { clearDatabase, closeWindow, actionFiresTopicTest } = require("spec/util/TestUtils");
+var { clearDatabase, closeWindow, actionFiresTopicTest, waitFor } = require("spec/util/TestUtils");
 var moment = require("lib/moment");
 
 describe("SampleHistory controller", function() {
@@ -38,10 +38,15 @@ describe("SampleHistory controller", function() {
     expect( result.sampleId ).to.exist;
   });
 
-  it('places a Sync button on the anchor bar toolbar', function() {
+  // The caption is on the button as soon as it is built; the label it announces
+  // itself by is not. iOS drops an accessibilityLabel written to a view it has
+  // not realised yet — a direct assignment is dropped just the same — so the
+  // binding's re-apply on first layout is what puts it there. Wait for the value
+  // the button actually carries once it is on screen.
+  it('places a Sync button on the anchor bar toolbar', async function() {
     expect( ctl.syncButton ).to.exist;
     expect( ctl.syncButton.label.text ).to.equal("SYNC");
-    expect( ctl.syncButton.label.accessibilityLabel ).to.equal("Sync");
+    await waitFor( () => ctl.syncButton.label.accessibilityLabel === "Sync" );
   });
 
   // The Sync button fires the START_SYNC intent; Main routes it to the


walta-app/app/controllers/NavButton.js
```diff
@@ -1,61 +1,71 @@
-var Topics = require('ui/Topics');
-var { disableControl, enableControl } = require("ui/ViewUtils");
+// Presenter shell only — the Titanium-free lib/mvvm/controllers/NavButton binds
+// the button's colours, caption and tap to a NavButtonViewModel that owns them.
+// What is left here is the glue bindView cannot express: an icon view that does
+// not exist until a screen asks for one, and the setter API the GoBack/GoForward/
+// Assess buttons inherit through Alloy's baseController.
+// See docs/patterns/screen-controllers.md.
+var NavButtonViewModel = require('mvvm/viewmodels/NavButton');
```
TheCodeSharman: waht do we need to do to move the NavButton and AnchorBar and TopLevelWindow inside the MVVM pattern entirely, this file should be empty.

TheCodeSharman: I'm thinking the nav buttons shold be child components of an AnchorBar view that is created by TopLevelWindow it it needs a planning session.

TheCodeSharman: Lets create a create to come up with the TopLevelWindow and AnchorBar MVVM pattern and merge this PR, we can address the design issue later.

TheCodeSharman: Carded as [WB-322 — Design the TopLevelWindow + AnchorBar MVVM pattern](https://trello.com/c/hxgYJ24m), in the Architecture/Infrastructure backlog.

You're right that this file should be empty. Three things keep it alive, and none of them can go until the bar itself moves:

1. **NavButton is never built through the View seam.** `View.createComponent` is what pairs an Alloy controller with its `lib/mvvm/controllers/<name>`. NavButton is built by `Alloy.createController("NavButton")` in SampleHistory/Summary, and by Alloy `baseController` inheritance in GoBack/GoForward/Assess — so the shell has to construct its own view-model and binder, doing View's job.
2. **The icon view doesn't exist at bind time** — `setIconLeft`/`setIconRight` create it on demand, so it can't sit in a static bindings map.
3. **`cancelBubble`** — Titanium event semantics bindView doesn't model.

Your direction is on the card as the one to evaluate: nav buttons as child components of an AnchorBar view created by TopLevelWindow, mounted through `collection`/`component` so createComponent attaches each button's controller and the shells disappear. That also picks up the bar's icon tools, which `createToolBarButton` still builds as bare Ti views and which therefore *don't* get a pressed state from this PR.

I've noted the open questions there — what replaces `baseController` for the three derived buttons, whether the icon becomes a declared-and-hidden child, and whether bindView grows `cancelBubble` or the restructure makes it moot. Also flagged that the WB-314 anchor-bar layout constraints belong in that redesign rather than as a patch.

PR #502, opened 2026-09-02
WB-309: Android: SampleTray scrollTo(0) is ignored, hanging the tray scroll specs

TheCodeSharman:
## Summary

**The tray could open on Android showing a blank grey panel.** The reveal's `scrollTo` is issued while Titanium is still laying the tray out, so it is clamped to the width the tray still has — and Titanium then reports the offset it was *asked for* as though it had moved. The view-model believes it is at the right edge and windows out every visible tile.

Proof, not inference: a screenshot taken while the spec hung shows the tray at offset 0 (endcap on the left, rest blank) while `contentOffset` reported 2505.

- **The tray reports its own laid-out width back** through a `measure` binding. That is what tells the model the offset it wants is reachable, and it replaces the per-postlayout retry that was compensating for the same thing by accident.
- **The scroll target is the difference of two measured widths**, not the dip→system conversion of their difference — that conversion landed a pixel past what the ScrollView could reach.
- **The reveal is an outstanding intent, not a one-shot.** Since Titanium's reported offset cannot distinguish a landed scroll from a clamped one, the model keeps asking on each layout pass and stops only when the tray reports an offset it did not ask for — the one piece of evidence that a finger has taken over. CI's own trace is what settled this: asking once landed on the pass that gets clamped, and skipped the two later passes that would have worked.
- **A layout pass that moves nothing no longer re-reveals**, so a tray scrolled left stays where the user put it — the yank-back this card was originally about.
- **New `deferredCommand` marker in `bindView`**, for a widget call Titanium ignores in the frame of the layout that provoked it (measured: a `scrollTo` dropped at +5 ms, accepted at +200 ms). This is a Titanium fact, so it lives in the presentation DSL rather than the view-model — per CLAUDE.md "grow `bindView` to capture Titanium rather than escape it". **Worth a look on review.**
- **Drive-by:** the `makeBinder` marker test hand-listed three markers, which is why a missing `binder.deferredCommand` passed Node and only failed on device. It now asserts the binder carries *every* marker the module exports.

Fixes [Trello WB-309](https://trello.com/c/YCxAg64q). Unblocks #499, whose only red leg was this spec.

## Test plan
- [x] `npx grunt unit-test-node` — 799 passing
- [x] `npx grunt --platform=android --simulator unit-test` (clean build) — 502 passing, 0 failing; the same suite is 502/1 on `main`
- [x] Tray specs on an **API 34** emulator built to match CI — 48 passing
- [x] Tray specs on API 36 — 48 passing
- [x] Full CI — all 16 jobs green

### DIFF
--- a/walta-app/app/lib/mvvm/viewmodels/IceCubeTray.js
+++ b/walta-app/app/lib/mvvm/viewmodels/IceCubeTray.js
@@ -3,10 +3,6 @@ const { endcapTile, interiorTile } = require("./SampleTrayTile");
 
 const identity = (x) => x;
 
-// Dip-to-system rounding across the geometry chain leaves the reported offset a
-// pixel or so either side of the one that was asked for.
-const OFFSET_TOLERANCE = 2;
-
 // The tile + slot VMs (SampleTrayTile — endcap or interior — and its
 // SampleTaxaIcon/SampleTrayPlus slots) read their geometry, kind and content
 // back through the public accessors here — but they're handed `owner` (which
@@ -23,7 +19,6 @@ class IceCubeTrayViewModel extends ChangeNotifier {
     this._viewportHeight = 0;
     this._viewportWidthPx = 0;
     this._trayWidthPx = 0;
-    this._revealPending = true;   // the tray opens showing its newest taxon
     this._scrollx = 0;
     this._tileCache = new Map();
     // Lazy, not built here: an owner (e.g. SampleTrayViewModel) constructs its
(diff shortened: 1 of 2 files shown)


walta-app/app/lib/mvvm/viewmodels/IceCubeTray.js
```diff
@@ -37,13 +44,35 @@ class IceCubeTrayViewModel extends ChangeNotifier {
...
+
+  // Revealing the right edge is an intent, not a one-shot. Titanium reports the
+  // offset it was *asked* for whether or not the scroll landed, so the model can
+  // never tell a landed scroll from a clamped one and cannot know how many layout
+  // passes the tray will take to reach the width it has been given. So it keeps
+  // asking on each pass — and stops the moment the tray is somewhere it did not
+  // ask for, which is the only evidence it gets that a finger has taken over.
+  _askForRightEdge() {
+    if (this._revealPending) this.trigger("scrollToRightEnd");
   }
```
TheCodeSharman: Is this titanium hacks leaking into the view model - the comments suggest it is?

TheCodeSharman: Yes — partly, and you've put your finger on the right line. Splitting what's here:

- `setTrayWidth` / `scrollTargetX` as the difference of two *measured* widths — I'd argue this isn't a leak. The model is told a measured fact and derives geometry from it, the same shape as the `setViewport` that was already here.
- `_askForRightEdge` keeping the ask outstanding across layout passes — **this is the leak.** Its entire justification is that Titanium clamps the scroll and then reports the offset it was asked for. Nothing in the product says "re-ask on every layout pass".
- `setScrollOffset` clearing `_revealPending` — mixed. "Stop revealing once a finger has taken over" is the behaviour this card is about; comparing against the target within a 2px tolerance is Titanium rounding.

The part I missed: `04cbfeef` (*let command hold itself back*) landed **after** the retry, and it puts this exact quirk in `bindView`'s implementation — a command raised inside a layout callback now defers itself out of that frame. That may make the view-model retry redundant. If it is, the model collapses to product logic only — *reveal the right edge when the tray gets a new width, or when the taxa change* — and `_revealPending`, `OFFSET_TOLERANCE` and the `setViewport` trigger all go (rotation still reveals, transitively: the relayout reports a new width).

Testing that now. The original failure only ever reproduced on CI's 640x288 viewport, so CI is the arbiter — if it stays green the leak deletes itself; if not, the retry is load-bearing and belongs inside `bindView`, not here.

PR #525, opened 2026-09-11
WB-341: Persist the earned belt level to the user's CERDI account

TheCodeSharman:
## Summary
- A belt was a local record only — `BeltRepository.js` said so in its header — so it was lost on reinstall and did not follow the account to a second device. It now reaches CERDI as `belt_level` and is read back as `qaqc_level`, using the `retrieveUser()` / `updateUser()` methods added in #523.
- **Offline-first.** Training happens without a network, so the local award stays authoritative and immediate; the send is deferred and retried. Triggers: finishing a session, signing in, the network returning, and app start.
- **Pending is a comparison, not a flag.** New `syncedBeltLevel` column beside `beltLevel` (migration `202609110000000_user.js`). `beltNeedingPush()` derives pending from the two, so there is no second record to drift — and `syncedBeltLevel` is only written once the request resolves, which is the mistake WB-101 made with `serverSyncTime`.
- **A push reads the server before it writes.** This was a hole in the original plan, caught by a test: a belt earned offline on one device would otherwise demote an account another device had already promoted, because the pending level is lower and the server takes what it is sent. Test: *"does not demote the server to a belt earned offline here"*.
- **Signing in** claims any signed-out belt, then reads the account's level and keeps the higher — so a fresh install wears a belt earned elsewhere, and neither side silently demotes the other.

### Notes for review
- `setLevel` moved from `INSERT OR REPLACE` to an upsert. `REPLACE` deletes and re-inserts the row, which silently dropped `syncedBeltLevel`. It happened to leave the belt pending, which is the right outcome by accident — the upsert makes it the outcome on purpose.
- **The wiring in `index-app.js` has no test.** It is the composition root and there is no spec for it — the pre-existing `claimAnonymousBelt` subscription there is equally uncovered. Every unit behind it is tested; the glue is not. Worth a reviewer's eye.
- **Deviation from the plan on the card:** the design said to add a branch to `SampleSync.resumeInterruptedWork`. I did not. Once the push reconciles on its own it no longer needs the sync's pending-work machinery, and keeping belts out of `SampleSync` avoids coupling sample sync to training — `SampleSync.init()` also runs ~80 lines before `beltAwards` exists, so injecting it would have meant reordering startup.
- **Still open, not in this PR:** the Menu reads the belt once at controller construction, so a level *pulled* from the server after login has no path to the view. A fresh install will show the belt on the next launch, not the moment it signs in.

Implements [Trello WB-341](https://trello.com/c/YsDusnzi/341-persist-the-earned-belt-level-to-the-user-s-cerdi-account); the contract it relies on was pinned in #523.

## Test plan
- [x] `npx grunt unit-test-node` — 894 passing (22 in `BeltAwards`, incl. offline, failure-leaves-pending, and demotion-protection)
- [x] `npx grunt --platform=ios --simulator unit-test` — 552 passing, 0 failing (18 in `BeltRepository`, incl. persistence across close+reopen)
- [x] Android emulator unit tests — CI green (run 34564835957)
- [x] Acceptance + e2e, both platforms — CI green (16/16)
- [ ] Manual: earn a belt in airplane mode, restore network, confirm it reaches the account — **not done**. This is the one path CI cannot reach: the `index-app.js` wiring is the uncovered part, and only a real device with the radio off exercises it end to end.


test/logic/BeltAwards_spec.js
```diff
@@ -51,6 +51,109 @@ describe("logic/BeltAwards", function () {
     expect(awards.currentBelt()).to.equal(null);
   });
 
+  it("pushes a newly earned belt to the server", async function () {
+    const { awards, cerdiApi } = make();
+    awards.awardFor("101");
+    await awards.pushPendingBelt();
+    expect(cerdiApi.updates).to.deep.equal([{ belt_level: 3 }]);
```
TheCodeSharman: I think for consistency we should use the canonical name rather than the alias

```suggestion
    expect(cerdiApi.updates).to.deep.equal([{ qaqc_level: 3 }]);
```

TheCodeSharman: Done — and good call. I'd assumed `belt_level` was the only accepted write name; the sandbox's own OpenAPI spec lists `qaqc_level` as a first-class property of the `PUT /user` request body, with `belt_level` described as "Alias for `qaqc_level` accepted on submission only".

Confirmed against the live sandbox rather than trusting the spec — the contract tests now write `qaqc_level` and read it back, all four green. `BeltAwards` sends `{ qaqc_level: level }`, the unit specs assert it, and nothing in the app references the alias any more, so CERDI can retire it whenever they like.

826ca6f8

PR #527, opened 2026-09-12
WB-343: "Which question did I get wrong?" names the wrong couplet

TheCodeSharman:
## Summary
- **The bug, in the reporter's words:** correct answer *damselfly*, learner answered *Baetid mayfly*, and the hint said couplet `node_15` — "you chose 'with obvious tails', should have been 'no obvious tails'". Damselfly larvae **do** have tails, so the advice contradicts the key.
- **Why.** The key is a **DAG, not a tree** (see the diagram from #526). Eight nodes have more than one parent, so **47 of 156 taxa sit at the end of more than one route**. Each node records a single `parentLink`, and *which* one is a loader accident — `KeyLoaderXml` keeps the last parent (`// will clobber previous .. not sure what to do about this`), `KeyLoaderInk` the first. Here the mis-routed side was the **expected** taxon: a damselfly is reachable both via `order_odonata` (the "no obvious tails" branch `parentLink` holds) and via `old_mayfly_start_point` (the "with obvious tails" branch the learner walked).
- **The fix.** Enumerate every route to both taxa and take the pair that parts **latest** — the couplet where the reader actually had to tell the two animals apart. The reported case now hints `old_mayfly_start_point`: *"you chose 'tails thin, round in cross section', should have been 'tails flattened, leaf-like or broad, jaw large and folded away under head'"*. Both animals have tails at that point, which is the whole reason the couplet exists.
- **Folded in:** the hint is now looked up by **taxonId rather than ref**, which settles one creature sitting at two places in the key. The coarse "order level ID" stop and the family below it share a taxonId (`bivalvia`/`hyriidae`, taxonId 179 — expected by shipped exercise `999`), and `attachTaxon` let only one survive in the index. Either is a legitimate answer, so the hint takes whichever of the two places parts *latest* from the reader's own route — the one they got nearest to. It is not a preference for the shorter route; length never enters the comparison.
- **Not indexed, enumerated per call — deliberately.** An earlier design built a multi-parent index in `linkNodeToParent`. That would have been **empty in the app**: neither production loader calls it, both rehydrate pre-wired outcomes via `attachNode`/`attachTaxon`. It would have passed every hand-built test and fixed nothing. Enumeration is 220 routes over 156 taxa on a once-per-tap action.
- `findIncorrectDecision` had no callers and goes with it. The taxonomy diagram's wording follows the removal of `pathFromRoot` (it named a function this PR deletes) — hence the regenerated SVG in the diff.

[Trello WB-343](https://trello.com/c/WjkPUEQ8/343-which-question-did-i-get-wrong-names-the-wrong-couplet) — **the hint algorithm only**.

### Deliberately out of scope
- **Capturing the route the learner actually walked.** Not needed for this defect (the selected side had only one route), but of the 7,392 pairs whose selected taxon is multi-route, **1,987 (27%)** get a different couplet depending which way it was walked. Needs a `training_taxa` migration and threading — its own card.
- **Declaring the canonical link + failing the build on ≠1.** Hygiene rather than correctness: measured, the *best possible* canonical assignment still mismatches 1,710 of 24,180 pairs, and fixing `k_odonata` this way makes 880 other pairs worse. Worth doing anyway — it also repairs `goUp()`, which walks `parentLink` and so can drop a learner into a branch they never came from.

## Test plan
- [x] `npx grunt unit-test-node` — 885 passing
- [x] New `test/logic/Key_hint_realkey_spec.js` drives the **real shipped key**, red before the fix: asserts the damselfly/baetid pair names `old_mayfly_start_point`
- [x] Characterisation tests pinning what must not move: the limpet/mussel pair the acceptance scenario walks still parts at `k_mollusca`; the duplicate taxonId resolves to the deeper position
- [x] Multi-parent fixture tests — a node given two parents, asserting the couplet on the reader's own route rather than the one `parentLink` holds, in both directions
- [x] `test/logic/fixtures/key_walk_snapshot.txt` **unchanged** — confirms no taxon index was touched
- [ ] `npx grunt --platform=ios acceptance-test` — `features/academy.feature` "correcting a mistake"
- [ ] Device specs for the tick/cross rendering (`KeySearch_spec`, `QuestionController_spec`)
- [ ] Manual: exercise `101` on the simulator, answer damselfly with a Baetid, confirm the hint lands on the tails-shape couplet


TheCodeSharman:
Was this PR supposed to include the key graph diagram

PR #528, opened 2026-09-12
WB-344: Hint from the route the learner actually walked

TheCodeSharman:
> **Stacked on #527** — base is `task/wb-343-hint-couplet`, not `main`, so Files Changed shows only this change. Retarget to `main` once #527 merges.

## Summary
- #527 fixed the reported damselfly bug by enumerating routes and taking the pair that parts latest. That is still a **guess about how the reader got there** — and where a taxon can be reached more than one way the guess can name a question they were never asked. Of the pairs whose answer has more than one route, **1,987 of 7,392 (27%)** part at a different couplet depending which way it was walked. Maggots are the clearest: reachable from either side of the worm-like couplet, so the same wrong answer has two different right explanations.
- **Record the couplets walked, and hint from those.** The named couplet is then always one the reader actually answered — verified across all 34,098 taxon-pair cases: **0** name a couplet off the walked trail, 0 produce no hint, 0 diverge at index 0.
- **Where the trail lives.** A `logic/KeyTrail.js` service built at start-up beside the key it reads — *not* on `Key` (one object shared with the survey flow, whose single mutable cursor is already a wart) and not a global. Nodes are held by identity, not ref: the Ink loader can mint anonymous nodes with an empty id.
- **Recording it is two lines** in the key-search view-model, because every screen open already re-anchors the key there. That is what makes going up, going back and jumping land on the right prefix without any of them needing separate handling: `enter()` rewinds to a couplet already walked, and starts afresh on one that wasn't.
- **The stored route is a filter, not a path to resolve.** It selects from the routes the key can still trace, so a route written before a taxonomy edit selects nothing and falls back to #527's answer. No validation code, no corrupt-route branch, and rows written before the column existed take the same path.
- **Correcting an answer** re-enters the key mid-couplet with no walk above it, so the hint now carries the route that reached it and the corrected walk continues from there rather than starting blind.

[Trello WB-344](https://trello.com/c/kTdkW3Nu/344-hint-from-the-route-the-learner-actually-walked)

### Flagged for your call
- **`goUp()` is still broken and this PR does not fix it.** It navigates by `parentLink`, so from one of the eight convergence nodes it drops the reader into a branch they never came from. `trail.parentOf(node)` is now available and is the real fix — ~2 lines plus a spec — but it changes **survey** navigation too, not just training, so I left it out rather than assume. Worth doing in its own PR.
- **`Training` reads the injected trail** rather than the route being threaded through the `IDENTIFY` payload. The alternative would make `TaxonDetails` (an Alloy shell) read navigational state and put it on the bus, which the MVVM rule forbids. It does mean one piece of ambient state, resting on the invariant that *in training the key is the only path to an identification* — enforced in three places (`MethodSelect`, `KeySearch`, `TaxonDetails`).
- **The trail defaults to one that records nothing**, so a screen wired without it still works — and would silently stop recording. `test/controllers/KeySearch_spec.js` exists only to catch that; don't drop it.

## Test plan
- [x] `npx grunt unit-test-node` — 910 passing
- [x] `npx grunt build-test` — 302 passing
- [x] Real-key tests for the maggot case: the same wrong answer hints `node_89` for the reader who came via `node_92` and `node_91` for the one who came via `node_91`
- [x] A stale route that no longer traces the key falls back rather than failing
- [x] `KeyTrail` unit spec: records, rewinds on re-entry, restarts off-trail, seed beats rewind
- [x] Existing hint contract tests updated for the added `route` field
- [ ] Device: `TrainingRepository_spec` route round-trip through close/reopen, plus a row stored without one — **written, not yet run on a device**
- [ ] `features/academy.feature` "correcting a mistake" — the seeded re-entry is exactly what it walks
- [ ] Manual: exercise `101`, identify a maggot both ways, confirm the two hints differ


TheCodeSharman:
OK there's a problem here: the hint MUST be the route the user took DO NOT chose the shorter path if there are multiple. The reason is that this is trying to identify the *specifc* couplet the user answers incorrectly for learning purposs.

test/logic/KeyTrail_spec.js
```diff
@@ -0,0 +1,127 @@
...
+//
+//              n1 (root)
+//            /          \
+//          n2            n3
+//         /  \          /  \
+//       t1    shared <-'    t4
+//             /    \
+//           t2      t3
+//
+function buildKey() {
```
TheCodeSharman: is this something thtat belongs in a fixture helper class?

walta-app/app/lib/mvvm/viewmodels/KeySearch.js
```diff
@@ -8,19 +9,24 @@ const QuestionViewModel = require("./Question");
 // nowhere else — navigate on and the next screen builds without it, come back
 // and it is applied again. Titanium-free.
 class KeySearchViewModel extends ChangeNotifier {
-  constructor({ key, node, topics, hint = null, surveyType = null, allowAddToSample = false, position = null, training = false }) {
+  constructor({ key, node, topics, hint = null, trail = KeyTrail.none, surveyType = null, allowAddToSample = false, position = null, training = false }) {
```
TheCodeSharman: ok I think there is a single responsibility problem here - the KeyTrail is entirely training module concern, but its being passed into the KeySearch controller is is supposed to be generic. The caller should pass the correct and incorrect couplet for the hint, but the hint decsion should be made by the training module not here. So why does the KeySearch depend on a KeyTrail object??

TheCodeSharman: Tried it and backed it out — agreed it isn't worth the cost.

For the record, the hint *decision* was already outside this screen: `KeySearch` receives `{nodeId, correctRef, incorrectRef}` and only renders verdicts from it; `TaxonComparison.whichQuestion()` is what decides. What #528 added here is the recording, which is the part you're pointing at.

The alternative was to keep the trail as a subscriber to a `KEY_POSITION_OPENED` event fired from `updateDecisionWindow` in Main.js — the one funnel every route into the key passes through. It works (the trail and its specs both rewrote cleanly, plus a gate so survey walks record nothing), but it needs a `via:` marker threaded through all four router call sites so the trail can tell walking on from jumping in, and it moves the recording a long way from the thing being recorded. More moving parts for a screen that is only `enter`/`step`.

So the trail stays a `KeySearch` parameter. No change on the branch.

TheCodeSharman: We looked at the alternatives, and to move KeyTrail out would require subscribing to events and the reaulting code less coherent. So I think KeySearch is the right place. Its not making the decisions here its just tracking the history.

PR #536, opened 2026-09-22
WB-357: Cover the creature-photo tap on the taxon comparison screen

TheCodeSharman:
## Summary

Raised as a suspected iOS defect while restructuring this view in #534. **It is not one — the tap works.** This PR lands the evidence instead of a fix.

- **What was suspected.** `TaxonComparisonPhoto` carries its click listener on the root container while the `photo` ImageView filling the card has none. That is the exact shape of the `SampleTaxaIcon` defect fixed in PR #385: Titanium on iOS raises no bubbling `click` from a listener-less hit-test target, so a tap on the creature did nothing there.
- **What actually happens.** Driven through Appium against the real modal, the tap reaches the handler and the taxon's detail screen opens — iOS as well as Android. So the structural resemblance alone doesn't predict the fault. The likeliest difference is that this container carries a bound `accessibilityLabel`, making it an accessibility element in its own right. **No production code is changed by this PR.**
- **Why it had never been checked.** No cheaper layer can see it. A device spec fires `click` synthetically on the container, which bubbles on both platforms and passes either way — so it cannot distinguish a reachable listener from an unreachable one. Only a real hit-tested tap can, and nothing tapped a comparison photo: the screen object only clicked the action button.

So the gap this closes is real even though the defect wasn't: a user path with no coverage at the one layer that can cover it, now guarded.

### Why a separate scenario

Splicing the tap into the existing correction scenario doesn't work: browsing out to a taxon and coming back leaves the tray in a state where the comparison **no longer re-opens** (`I select the incorrect taxon` then fails). That may be intended — browsing out is leaving the feedback — or it may be worth a look on its own. Either way it's a separate question, so this scenario stands alone and ends at its assertion rather than dragging that behaviour into an unrelated test.

### Two harness fixes that came with it

The scenario passed alone and failed in the suite, on iOS only, waiting for `~Cell 1.` — a label only an **empty** tray slot carries. Both causes are fixed here, because the scenario is red without them:

- **`AppReset` didn't clear training state** (WB-357). Between scenarios iOS does a soft in-app reset rather than Android's `pm clear`, and that reset wipes samples, taxa, tokens and logs but not training. A training session is deliberately durable so an app the OS reclaims resumes where it left off — which also means it outlives the scenario that started it, and the academy then resumes a finished exercise instead of starting a fresh one. Reset now clears the training repository alongside the log one.
- **`Then an empty training tray is shown` asserted nothing** ([WB-358](https://trello.com/c/HpHQqlDY)). It only waited for the tray *screen*, so "empty" was a claim the step never made — which is why the leak surfaced three steps downstream as a locator timeout naming the wrong thing. It now asserts both halves of empty, without needing the exercise's creature count: the first slot is still a numbered placeholder, and nothing has been identified into the tray at all.

Both were verified by backing the fix out and watching the failure land where it should.

Fixes [Trello WB-357](https://trello.com/c/KJjn04oV) — card updated to record the disproof rather than the original theory — and [WB-358](https://trello.com/c/HpHQqlDY).

## Test plan
- [x] `npx grunt --platform=ios --simulator acceptance-test --grep="Browse to a creature from the training feedback"` — 14 steps, all passed
- [x] `npx grunt --platform=android --simulator acceptance-test --grep="Browse to a creature from the training feedback"` — 14 steps, all passed
- [x] The pre-existing correction scenario left byte-identical
- [x] Both academy scenarios together on iOS (the ordering that failed in CI) — 35 steps, all passed
- [x] Empty-tray step verified red-then-green: with the reset fix backed out it fails at the step making the claim ("Cell 1 not present"), not three steps later
- [ ] CI green


features/academy.feature
```diff
@@ -21,3 +21,19 @@ Scenario: Complete a training exercise, correcting a mistake
   When I finish the training
   Then the menu is shown
   And I am wearing a "Yellow belt with a white tip"
+
+# A creature on the feedback is a way into the key as well as an explanation.
+# This lives at the acceptance layer because only a real hit-tested tap can show
+# the listener is reachable: a device spec fires `click` synthetically on the
+# container, which bubbles on both platforms and so passes either way.
```
TheCodeSharman:
```suggestion

```

PR #542, opened 2026-09-22
WB-354: Implement delete user account functionality

TheCodeSharman:
## Summary
- Delete Account opens a modal that says plainly what will happen — the account goes, the submitted samples are anonymised — and asks for the password before anything is destroyed.
- **The password is proved against the server first**, so a mistyped one costs nothing: only once `/token/create` accepts it is `DELETE /user` called, and only then is the app signed out and sent home.
- A wrong password and a failed delete both leave the account exactly as it was, say so through the dialog seam, and leave the modal up to try again.
- Saves the CERDI API spec the card attached to `walta-doc/api-server/api.yaml`. (The card said `walta-docs/`; the folder in this repo is `walta-doc/`.)

### The bug this found
`DELETE /user` answers **204 No Content**, and it is the first call in the app that does. Parsing that empty body threw **inside Titanium's own HTTP callback**, where no caller can catch it — so the request never settled and the app sat on a promise that would never resolve. **Deleting an account left the app wedged rather than returning to the menu.**

No unit test would have caught this: the throw is in the HTTP callback, not in anything the view-model calls. The acceptance scenario is what surfaced it, and `sendOnce` now treats an empty body as no content.

Two smaller things the same scenario forced out, both real rather than test-only:
- The modal now sits in a **ScrollView** (the treatment the entry screens get) — without it the keyboard covered both buttons.
- The warning label **no longer carries an accessibility label**. It was `delete_account_warning`, which is what a screen reader would have announced *instead of the warning itself*.

## Test plan
- [x] `npx grunt unit-test-node` — 949 passing
- [x] iOS device specs `DeleteAccount`, `Account`, visual manifest — 58 passing, layout checked against the mock-up
- [x] iOS acceptance `features/account.feature` — 2 scenarios, 19 steps, all passing
- [ ] Visual check on Android emulator

### Not covered at the acceptance layer
The **wrong-password** path is covered by the unit and device specs but not by a cucumber scenario: the message is a native alert, and while one is up the driver cannot see the app behind it to assert on what was left alone. Flagging rather than leaving a scenario that looks like coverage but is fighting the driver.

[Trello WB-354](https://trello.com/c/Nc6pqMgX/354-implement-delete-user-account-functionality).

**Based on `task/wb-353-account-details` ([#541](https://github.com/TheWaterBugCompany/WALTA/pull/541))**, itself on #540 → #539 → #538. Retarget as they merge.


walta-app/app/lib/mvvm/viewmodels/DeleteAccount.js
```diff
@@ -0,0 +1,48 @@
...
+    if (next === this._password) return;
+    this._password = next;
+    this.notifyListeners();
+  }
+
+  get deleteEnabled() { return this._password.length > 0; }
+
+  confirmDelete() {
+    if (!this.deleteEnabled) return Promise.resolve();
+    return this._cerdiApi.loginUser(this._cerdiApi.retrieveUsername(), this._password)
```
TheCodeSharman: If loginUser fails is the user left logged out? We want the user to remain loggin in even if they get ehri apssword wrong here. Its just a way to gate accidental deletes from fat fingers.

TheCodeSharman: No — a wrong password leaves the user signed in.

`loginUser` only calls `storeUserToken` inside its success `.then`, so a rejection never touches the stored token. The only places that clear it are `AppReset`, an explicit log out, and `_accountIsGone` after the delete actually succeeds. The 401 handling in `withServerAccessToken` refreshes the *server* token, not the user's.

The view model already had `keeps the account signed in when the password is wrong`, but that drives a fake whose `loginUser` simply rejects — it proves the view model behaves and says nothing about the client underneath it, which is exactly the half your question was about. Added `leaves the signed-in user alone when the password is wrong` against the real `CerdiApi` over a scripted 401, asserting both the stored token and the signed-in username survive.

One wrinkle worth knowing, not a logout: a wrong *user* password comes back 401, and `withServerAccessToken` can't tell that from a stale *server* token, so it evicts a perfectly good server token and retries once before failing. Harmless but wasteful — say the word and I'll raise a card.

